PR #16, opened 2019-11-08
Add various shutil methods

martinhoefling:
Added various methods like copy2, copytree etc. Also added a kwarg that allows to perform copying server side. Many methods are copied and adapted from their python2 shutil counterpart to mimic their behavior. Further, a set of integration tests that at least cover the happy paths was added. @jborean93 let me know what you think.

### DIFF
--- a/smbclient/_os.py
+++ b/smbclient/_os.py
@@ -55,7 +55,7 @@ from smbprotocol.file_info import (
 from smbprotocol.ioctl import (
     CtlCode,
     IOCTLFlags,
-)
+    SMB2SrvCopyChunk, SMB2SrvCopyChunkResponse, SMB2SrvCopyChunkCopy, SMB2SrvRequestResumeKey)
 
 from smbprotocol.open import (
     CreateOptions,
@@ -75,10 +75,11 @@ from smbprotocol.structure import (
     DateTimeField,
 )
 
-
 XATTR_CREATE = getattr(os, 'XATTR_CREATE', 1)
 XATTR_REPLACE = getattr(os, 'XATTR_REPLACE', 2)
 
+MAX_COPY_CHUNK_SIZE = 1 * 1024 * 1024  # maximum chunksize 1M from 3.3.3 in MS-SMB documentation
+MAX_COPY_CHUNK_COUNT = 16  # maximum total chunksize 16M from 3.3.3 in MS-SMB documentation
 
 SMBStatResult = collections.namedtuple('SMBStatResult', [
     'st_mode',
@@ -1016,3 +1017,93 @@ class SMBDirEntry(object):
             if not self._lstat:
                 self._lstat = lstat(self._smb_raw.name)
             return self._lstat
+
+
+def copyfile_server_side(src, dst, **kwargs):
+    """
+    Server side copy of a file
+
+    :param src: The source file.
+    :param dst: The target file.
+    :param kwargs: Common SMB Session arguments for smbclient.
+    """
+
+    if not _is_sameshare(src, dst):
+        raise ValueError(
+            "Server side copy can only occur on the same drive, '%s' must be the same as the dst root '%s'" % (
+                src, dst))
+
+    norm_dst = ntpath.normpath(dst)
+    norm_src = ntpath.normpath(src)
+
+    try:
+        stat(norm_dst, **kwargs)
+    except SMBOSError as err:
+        if err.errno != errno.ENOENT:
+            raise
+    else:
+        raise ValueError("Target %s already exists" % norm_dst)
+
+    with SMBRawIO(norm_src, mode='r', desired_access=FilePipePrinterAccessMask.GENERIC_READ,
+                  create_options=(CreateOptions.FILE_NON_DIRECTORY_FILE), share_access='r',
+                  **kwargs) as raw_src:
+
+        with SMBFileTransaction(raw_src) as transaction_src:
+            ioctl_request(transaction_src, CtlCode.FSCTL_SRV_REQUEST_RESUME_KEY,
+                          flags=IOCTLFlags.SMB2_0_IOCTL_IS_FSCTL,
+                          output_size=32)
+
+        val_resp = SMB2SrvRequestResumeKey()
+        val_resp.unpack(transaction_src.results[0])
+
+        chunks = _get_srv_copy_chunks(transaction_src)
+
+        with SMBRawIO(norm_dst, mode='x', desired_access=FilePipePrinterAccessMask.GENERIC_WRITE,
+                      share_access='r',
+                      create_options=CreateOptions.FILE_NON_DIRECTORY_FILE, **kwargs) as raw_dst:
+
+            with SMBFileTransaction(raw_dst) as transaction_dst:
+                for batch in _batches(chunks, MAX_COPY_CHUNK_COUNT):
+                    copychunkcopy_struct = SMB2SrvCopyChunkCopy()
+                    copychunkcopy_struct['source_key'] = val_resp['resume_key'].get_value()
+                    copychunkcopy_struct['chunks'] = batch
+
+                    ioctl_request(transaction_dst, CtlCode.FSCTL_SRV_COPYCHUNK_WRITE,
+                                  flags=IOCTLFlags.SMB2_0_IOCTL_IS_FSCTL,
+                                  output_size=32, input_buffer=copychunkcopy_struct)
+
+            for result in transaction_dst.results:
+                copychunk_response = SMB2SrvCopyChunkResponse()
+                copychunk_response.unpack(result)
+                if copychunk_response['chunks_written'].get_value() < 1:
+                    raise SMBOSError('Could not copy chunks in server side copy', filename=norm_dst)
+
+
+def _get_srv_copy_chunks(transaction):
+    chunks = []
+    offset = 0
+
+    while offset < transaction.raw.fd.end_of_file:
+        copychunk_struct = SMB2SrvCopyChunk()
+        copychunk_struct['source_offset'] = offset
+        copychunk_struct['target_offset'] = offset
+        if offset + MAX_COPY_CHUNK_SIZE < transaction.raw.fd.end_of_file:
+            copychunk_struct['length'] = MAX_COPY_CHUNK_SIZE
+        else:
+            copychunk_struct['length'] = transaction.raw.fd.end_of_file - offset
+
+        chunks.append(copychunk_struct)
+        offset += MAX_COPY_CHUNK_SIZE
+
+    return chunks
+
+
+def _batches(lst, n):
+    for i in range(0, len(lst), n):
+        yield lst[i:i + n]
+
+
+def _is_sameshare(src, dst):
+    unc_src, _ = ntpath.splitdrive(src)
+    unc_dst, _ = ntpath.splitdrive(dst)
+    return unc_src.lower() == unc_dst.lower()
--- a/smbclient/_os.py
+++ b/smbclient/_os.py
@@ -55,7 +55,11 @@ from smbprotocol.file_info import (
 from smbprotocol.ioctl import (
     CtlCode,
     IOCTLFlags,
-    SMB2SrvCopyChunk, SMB2SrvCopyChunkResponse, SMB2SrvCopyChunkCopy, SMB2SrvRequestResumeKey)
+    SMB2SrvCopyChunk,
+    SMB2SrvCopyChunkResponse,
+    SMB2SrvCopyChunkCopy,
+    SMB2SrvRequestResumeKey
+)
 
 from smbprotocol.open import (
     CreateOptions,
@@ -1019,7 +1023,7 @@ class SMBDirEntry(object):
             return self._lstat
 
 
-def copyfile_server_side(src, dst, **kwargs):
+def copyfile(src, dst, **kwargs):
     """
     Server side copy of a file
 
@@ -1062,8 +1066,8 @@ def copyfile_server_side(src, dst, **kwargs):
                       share_access='r',
                       create_options=CreateOptions.FILE_NON_DIRECTORY_FILE, **kwargs) as raw_dst:
 
-            with SMBFileTransaction(raw_dst) as transaction_dst:
-                for batch in _batches(chunks, MAX_COPY_CHUNK_COUNT):
+            for batch in _batches(chunks, MAX_COPY_CHUNK_COUNT):
+                with SMBFileTransaction(raw_dst) as transaction_dst:
                     copychunkcopy_struct = SMB2SrvCopyChunkCopy()
                     copychunkcopy_struct['source_key'] = val_resp['resume_key'].get_value()
                     copychunkcopy_struct['chunks'] = batch
@@ -1072,11 +1076,11 @@ def copyfile_server_side(src, dst, **kwargs):
                                   flags=IOCTLFlags.SMB2_0_IOCTL_IS_FSCTL,
                                   output_size=32, input_buffer=copychunkcopy_struct)
 
-            for result in transaction_dst.results:
-                copychunk_response = SMB2SrvCopyChunkResponse()
-                copychunk_response.unpack(result)
-                if copychunk_response['chunks_written'].get_value() < 1:
-                    raise SMBOSError('Could not copy chunks in server side copy', filename=norm_dst)
+                for result in transaction_dst.results:
+                    copychunk_response = SMB2SrvCopyChunkResponse()
+                    copychunk_response.unpack(result)
+                    if copychunk_response['chunks_written'].get_value() < 1:
+                        raise SMBOSError('Could not copy chunks in server side copy', filename=norm_dst)
 
 
 def _get_srv_copy_chunks(transaction):
--- a/smbclient/shutil.py
+++ b/smbclient/shutil.py
@@ -1,0 +1,306 @@
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2019, Jordan Borean (@jborean93) <jborean93@gmail.com>
+# MIT License (see LICENSE or https://opensource.org/licenses/MIT)
+from __future__ import unicode_literals
+import errno
+
+from ntpath import join, basename, normcase, abspath, normpath, splitdrive
+
+import sys
+import stat as py_stat
+from os import error as os_error
+
+from smbclient._io import ioctl_request, SMBFileTransaction, SMBRawIO
+from smbclient._os import remove, rmdir, listdir, lstat, stat, utime, open_file, makedirs, readlink, symlink
+from smbclient.path import islink, isdir
+from smbprotocol.exceptions import SMBOSError
+from smbprotocol.ioctl import SMB2SrvCopyChunk, IOCTLFlags, SMB2SrvCopyChunkResponse, CtlCode, SMB2SrvCopyChunkCopy, \
+    SMB2SrvRequestResumeKey
+from smbprotocol.open import FilePipePrinterAccessMask, CreateOptions, ShareAccess
+
+CHUNK_SIZE = 1 * 1024 * 1024  # maximum chunksize allowed by smb
+
+
+class Error(EnvironmentError):
+    pass
+
+
+class SpecialFileError(EnvironmentError):
+    """Raised when trying to do a kind of operation (e.g. copying) which is
+    not supported on a special file (e.g. a named pipe)"""
+
+
+def rmtree(path, ignore_errors=False, onerror=None, **kwargs):
+    """Recursively delete a directory tree.
+
+    If ignore_errors is set, errors are ignored; otherwise, if onerror
+    is set, it is called to handle the error with arguments (func,
+    path, exc_info) where func is smbclient.listdir, smbclient.remove,
+    or smbclient.rmdir;
+    path is the argument to that function that caused it to fail; and
+    exc_info is a tuple returned by sys.exc_info().  If ignore_errors
+    is false and onerror is None, an exception is raised.
+
+    :param path: The path to remove.
+    :param ignore_errors: The ignore errors flag.
+    :param onerror: The callback executed on errors
+    :param kwargs: Common arguments used to build the SMB Session.
+    :return: True if path is a dir or points to a dir.
+    """
+    if ignore_errors:
+        def onerror(*args):
+            pass
+    elif onerror is None:
+        def onerror(*args):
+            raise
+    try:
+        if islink(path, **kwargs):
+            # symlinks to directories are forbidden, see bug #1669
+            raise OSError("Cannot call rmtree on a symbolic link")
+    except OSError:
+        onerror(islink, path, sys.exc_info())
+        # can't continue even if onerror hook returns
+        return
+
+    names = []
+    try:
+        names = listdir(path, **kwargs)
+    except os_error:
+        onerror(listdir, path, sys.exc_info())
+    for name in names:
+        fullname = join(path, name)
+        try:
+            mode = lstat(fullname, **kwargs).st_mode
+        except os_error:
+            mode = 0
+        if py_stat.S_ISDIR(mode):
+            rmtree(fullname, ignore_errors, onerror, **kwargs)
+        else:
+            try:
+                remove(fullname, **kwargs)
+            except os_error:
+                onerror(remove, fullname, sys.exc_info())
+    try:
+        rmdir(path, **kwargs)
+    except os_error:
+        onerror(rmdir, path, sys.exc_info())
+
+
+def copy2(src, dst, server_side_copy=False, **kwargs):
+    """Copy data and metadata. Return the file's destination.
+
+    Metadata is copied with copystat(). Please see the copystat function
+    for more information.
+
+    The destination may be a directory.
+
+    """
+    if isdir(dst, **kwargs):
+        dst = join(dst, basename(src))
+    if server_side_copy:
+        copyfile_server_side(src, dst, **kwargs)
+    else:
+        copyfile(src, dst, **kwargs)
+    copystat(src, dst, **kwargs)
+
+
+def copystat(src, dst, **kwargs):
+    """Copy file metadata
+
+    Copy the permission bits, last access time, last modification time, and
+    flags from `src` to `dst`. The file contents, owner, and group are
+    unaffected. `src` and `dst` are path names given as strings.
+    """
+    st = stat(src, **kwargs)
+    utime(dst, ns=(st.st_atime_ns, st.st_mtime_ns), **kwargs)
+
+    # TODO: in principle, we can copy the readonly flag via chmod
+    # mode = py_stat.S_IMODE(st.st_mode)
+    # if hasattr(os, 'chmod'):
+    #     os.chmod(dst, mode)
+
+
+def copyfile(src, dst, **kwargs):
+    """Copy data from src to dst"""
+    if _samefile(src, dst):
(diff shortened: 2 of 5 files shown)


jborean93:
Awesome work thanks for doing this. I'm still working through some final stuff on the easy-api branch and once that is in I'll start the review on this so we can include it in the same release.

martinhoefling:
@jborean93 all right, I will try to fix the build in the meantime.

martinhoefling:
Some parts still require tests and I am not sure how to handle Py2/3 compatibility for some of the error formattings. I'll add some more tests for the missing branches.

martinhoefling:
Ah, this is actually a pytest bug when `pytest.raises` tries to match a unicode string. https://github.com/pytest-dev/pytest/issues/5478. I'll do the strict checking for the message only on python3.

smbclient/shutil.py
```diff
@@ -0,0 +1,306 @@
...
+
+from smbclient._io import ioctl_request, SMBFileTransaction, SMBRawIO
+from smbclient._os import remove, rmdir, listdir, lstat, stat, utime, open_file, makedirs, readlink, symlink
+from smbclient.path import islink, isdir
+from smbprotocol.exceptions import SMBOSError
+from smbprotocol.ioctl import SMB2SrvCopyChunk, IOCTLFlags, SMB2SrvCopyChunkResponse, CtlCode, SMB2SrvCopyChunkCopy, \
+    SMB2SrvRequestResumeKey
+from smbprotocol.open import FilePipePrinterAccessMask, CreateOptions, ShareAccess
+
+CHUNK_SIZE = 1 * 1024 * 1024  # maximum chunksize allowed by smb
```
jborean93: Where did you get this value from? SMB has a max packet size of around 65KB but newer dialects can go even larger than this.

smbclient/shutil.py
```diff
@@ -0,0 +1,306 @@
...
+    return chunks
+
+
+def _batches(lst, n):
+    for i in range(0, len(lst), n):
+        yield lst[i:i + n]
+
+
+def _samefile(src, dst):
+    return normcase(abspath(src)) == normcase(abspath(dst))
```
jborean93: If you want a proper samefile then `smbclient.path` has a function that actually compares the file id on the host but it is only really designed for remote -> remote comparisons.

smbclient/shutil.py
```diff
@@ -0,0 +1,306 @@
...
+    # TODO: in principle, we can copy the readonly flag via chmod
+    # mode = py_stat.S_IMODE(st.st_mode)
+    # if hasattr(os, 'chmod'):
+    #     os.chmod(dst, mode)
+
+
+def copyfile(src, dst, **kwargs):
+    """Copy data from src to dst"""
+    if _samefile(src, dst):
+        raise Error("`%s` and `%s` are the same file" % (src, dst))
```
jborean93: Better as `ValueError`

smbclient/shutil.py
```diff
@@ -0,0 +1,306 @@
...
+
+    for fn in [src, dst]:
+        try:
+            st = stat(fn, **kwargs)
+        except OSError:
+            # File most likely does not exist
+            pass
+        else:
+            # XXX What about other special files? (sockets, devices...)
+            if py_stat.S_ISFIFO(st.st_mode):
```
jborean93: We don't set this value at all, I've tried to find a way to determine the type of remote object, like a named pipe, outside of files/dirs/reparse points but have never really found a proper way for that.

smbclient/shutil.py
```diff
@@ -0,0 +1,306 @@
...
+                raise SpecialFileError("`%s` is a named pipe" % fn)
+
+    with open_file(src, 'rb', **kwargs) as fsrc:
+        with open_file(dst, 'wb', **kwargs) as fdst:
```
jborean93: `with` supports multiple entries so this could just be `with open_file(src, 'rb', **kwargs) as fsrc, open_file(dst, 'wb', **kwargs) as fdst:`. https://docs.python.org/3/reference/compound_stmts.html#the-with-statement

smbclient/shutil.py
```diff
@@ -0,0 +1,306 @@
...
+    """Copy file metadata
+
+    Copy the permission bits, last access time, last modification time, and
+    flags from `src` to `dst`. The file contents, owner, and group are
+    unaffected. `src` and `dst` are path names given as strings.
+    """
+    st = stat(src, **kwargs)
+    utime(dst, ns=(st.st_atime_ns, st.st_mtime_ns), **kwargs)
+
+    # TODO: in principle, we can copy the readonly flag via chmod
```
jborean93: This is partly why I don't think we should ship with the copy2 functionality for now. Better to focus on getting a basic file copy that supports all the permutations of local/remote for now.

jborean93:
Also, if you rebase the commit against easy-api I've removed Python 3.4 from the matrix as it is EOL and I've removed support for it.

smbclient/shutil.py
```diff
@@ -0,0 +1,306 @@
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2019, Jordan Borean (@jborean93) <jborean93@gmail.com>
```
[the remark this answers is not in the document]
martinhoefling: removed

smbclient/shutil.py
```diff
@@ -0,0 +1,306 @@
...
+    :return: True if path is a dir or points to a dir.
+    """
+    if ignore_errors:
+        def onerror(*args):
+            pass
+    elif onerror is None:
+        def onerror(*args):
+            raise
+    try:
+        if islink(path, **kwargs):
```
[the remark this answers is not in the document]
martinhoefling: removed the link check

smbclient/shutil.py
```diff
@@ -0,0 +1,306 @@
...
+            # symlinks to directories are forbidden, see bug #1669
+            raise OSError("Cannot call rmtree on a symbolic link")
+    except OSError:
+        onerror(islink, path, sys.exc_info())
+        # can't continue even if onerror hook returns
+        return
+
+    names = []
+    try:
+        names = listdir(path, **kwargs)
```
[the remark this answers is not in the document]
martinhoefling: I've written it according to your example in conftest with catching errors as rmtree does.

smbclient/shutil.py
```diff
@@ -0,0 +1,306 @@
...
+                remove(fullname, **kwargs)
+            except os_error:
+                onerror(remove, fullname, sys.exc_info())
+    try:
+        rmdir(path, **kwargs)
+    except os_error:
+        onerror(rmdir, path, sys.exc_info())
+
+
+def copy2(src, dst, server_side_copy=False, **kwargs):
```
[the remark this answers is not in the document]
martinhoefling: So what we're interested in is mainly server side copy of files and entire directory trees. Local <-> remote was not considered at all, your're right. We could add this, by checking the UNC paths and if it's a UNC path at all. If I understand your suggestion correctly, you still wouldn't call that "copy2", because of more things listed above, that need to be taken into consideration?
Maybe adding that points to the copy2 documentation is an alternative? The python documentation does it that way for example (https://docs.python.org/3/library/shutil.html). I also have no problem in renaming the functions to reserve the name for a more complete one in the future.

smbclient/shutil.py
```diff
@@ -0,0 +1,306 @@
...
+
+    try:
+        stat(norm_dst, **kwargs)
+    except SMBOSError as err:
+        if err.errno != errno.ENOENT:
+            raise
+    else:
+        raise ValueError("Target %s already exists" % norm_dst)
+
+    with SMBRawIO(norm_src, mode='r', desired_access=FilePipePrinterAccessMask.GENERIC_READ,
```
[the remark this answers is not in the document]
martinhoefling: After rebase, the test are failing, I'll have to find out why first.

smbclient/shutil.py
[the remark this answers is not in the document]
jborean93: I think we should have `copy` at a minimum and potentially look at `copyfileobj` snf `copyfile`. The `copy2`, `copymode`, and `copystat` should be reserved for the future when we have a clearer picture and more time to really think about how to implement it here. Due to the complexity I think it should be done in a separate PR and not as part of this basic implementation.

As for the local <-> remote question, I'm happy to just have remote <-> remote in this PR for now but remove the `server_side_copy` option. We can still raise a ValueError if the path isn't remote and just remove that in a future iteration. I don't think it warrants a `copy2` consideration, it still fits under the `copy` spec where it's just copying the file's metadata, smbclient just allows it to be done over SMB :).

smbclient/shutil.py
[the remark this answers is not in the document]
jborean93: Some of the failures were my doing sorry, I renamed some kwargs to avoid the builtin names. One I know off the top of my head is that `ioctl_request's` `buffer` kwarg is now `input_buffer`.

smbclient/shutil.py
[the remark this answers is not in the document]
martinhoefling: OK got it. Just the point with server_side_copy is not clear to me. Let's talk just about remote copy for a moment: Here, we can only do server_side_copy, if the server supports it (e.g. recent SAMBA versions) and we're on the same share/server. If not, we have to fall back to reading and writing chunks from python. Not sure how to find out if the server supports it though. That's why I've introduced the flag to indicate, that the server supports it (default is False).

The python chunk copy (which requires up and download of the files) can be easily reinforced to also work with local files.

smbclient/shutil.py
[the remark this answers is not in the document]
martinhoefling: Right, that was the issue!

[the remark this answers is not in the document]
martinhoefling: When changing to the suggested `open_file` I get
```
>       sid = self.raw.fd.tree_connect.session.session_id
E       AttributeError: '_io.TextIOWrapper' object has no attribute 'fd'
```

smbclient/shutil.py
martinhoefling: done

smbclient/shutil.py
martinhoefling: I removed the entire check.

smbclient/shutil.py
martinhoefling: I have extended samefile a bit to deal with certain scenarios.

smbclient/shutil.py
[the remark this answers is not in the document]
jborean93: It sounds like you need to open it in binary mode. Typically you will want to set `mode='<x>b` (the b is important) to make sure you are dealing with bytes. Internal functions should avoid using the text side unless they really need it.

smbclient/shutil.py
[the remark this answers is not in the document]
jborean93: Hmm ok I wasn't aware of that, looking at the Samba docs I can see that it introduced support for `FSCTL_SRV_COPYCHUNK` in 4.1.0 https://wiki.samba.org/index.php/Server-Side_Copy which has been EOL since 2016 https://wiki.samba.org/index.php/Samba_Release_Planning.

It's hard to tell when it was supported on the server side for Windows but considering

1. https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/eaa129d1-fca4-46b1-8ccd-b91f3e1d2d80#Appendix_A_Target_132 doesn't mention a minimum dialect version
2. https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb/c82ff91d-28d8-4274-a2ab-c2f5ade26b98 - It is supported in SMB1

I would say it has been supported on Windows for a while now.

Because of this I feel like we can rely on the fact that the server supports the server side copy operations and we don't need to fall back on our own manual read/write of chunks. If someone was using the ancient version of Samba I don't really want to make things easier for them (reason why I didn't add SMB1 support). As for whether it is still remote <-> remote but different server/share then we can just check the path to make sure they are the same before using the IOCTL commands or falling back to our own manual chunk read/writing.

Lastly, this isn't necessary but it sounds like the pure server/share copy through IOCTL should be a function in `smbclient._os` and `smbclient.shutil` just calls that if it's able to. The `._os` function is designed to expose the pure SMB side in an easy to use interface which this sounds like a good candidate. To better explain what I mean

```
# smbclient._os.py
def copy(src, dst, **kwargs):
    check that src and dst are on the same share
    Call IOCTL FSCTL_SRV_COPYCHUNK to copy the files


# smbclient.shutil.py
def copy(src, dst, **kwargs):
    if src and dst are on the same share
        call _os.py's copy(src, dst, **kwargs)
    else
        manually read and write from src/dst ourselves
```

smbclient/shutil.py
martinhoefling: This has nothing to do with the smb packet size (3.3.3 in the attached document). If I read this correctly, the widely used default is 1M per chunk max and max 16MB in chunks = 16 chunks of 1M. The 16 is hidden, I should extract it as variable as well.
[[MS-SMB].pdf](https://github.com/jborean93/smbprotocol/files/3833275/MS-SMB.pdf)

martinhoefling:
I have cleaned up the code a bit, rmtree, copy and copytree_copy are in shutil, the server side implementation has been moved to _os. I still wasn't able to make the SMBRawIO replacement working. I will check out next week.

smbclient/_os.py
```diff
@@ -55,7 +55,7 @@
 from smbprotocol.ioctl import (
     CtlCode,
     IOCTLFlags,
-)
+    SMB2SrvCopyChunk, SMB2SrvCopyChunkResponse, SMB2SrvCopyChunkCopy, SMB2SrvRequestResumeKey)
```
jborean93: Best to keep the same style as above and put an import on a newline. This is mostly to make it easier to alphabetise and add/remove imports without worrying about the line length

smbclient/_os.py
```diff
@@ -75,10 +75,11 @@
     DateTimeField,
 )
 
-
 XATTR_CREATE = getattr(os, 'XATTR_CREATE', 1)
 XATTR_REPLACE = getattr(os, 'XATTR_REPLACE', 2)
 
+MAX_COPY_CHUNK_SIZE = 1 * 1024 * 1024  # maximum chunksize 1M from 3.3.3 in MS-SMB documentation
```
jborean93: I'm still not sure where this size comes from? MS-SMB is the SMBv1 implementation which is a different protocol from SMBv2 (MS-SMB2). I also can't find mention of this in the docs on MS-SMB2 talking about receiving this request at https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/cd0162e4-7650-4293-8a2a-d696923203ef.

Maybe it's worth trying out a copy of a large file to see when it starts to fail to verify everything.

smbclient/_os.py
```diff
@@ -1016,3 +1017,93 @@ def stat(self, follow_symlinks=True):
             if not self._lstat:
                 self._lstat = lstat(self._smb_raw.name)
             return self._lstat
+
+
+def copyfile_server_side(src, dst, **kwargs):
```
jborean93: All the other methods in `_os.py` are all server side but I've kept the names as just `rename`, `replace`. I think we should follow that here and just have this as `copyfile` and make sure we add in the docstring that it only works for copying on the same server/share.

smbclient/shutil.py
```diff
@@ -0,0 +1,178 @@
...
+from smbclient._os import remove, rmdir, listdir, stat, makedirs, readlink, symlink, scandir, copyfile_server_side
+from smbclient.path import islink, isdir
+from smbprotocol.exceptions import SMBOSError
+
+
+class Error(EnvironmentError):
+    pass
+
+
+class SpecialFileError(EnvironmentError):
```
jborean93: Doesn't seem to be used anymore

smbclient/shutil.py
```diff
@@ -0,0 +1,178 @@
...
+    if _samefile(src, dst, **kwargs):
+        raise ValueError("`%s` and `%s` are the same file" % (src, dst))
+
+    if isdir(dst, **kwargs):
+        dst = join(dst, basename(src))
+
+    copyfile_server_side(src, dst, **kwargs)
+
+
+def copytree_copy(src, dst, symlinks=False, ignore=None, **kwargs):
```
jborean93: Should this just be `copytree`?

smbclient/shutil.py
```diff
@@ -0,0 +1,178 @@
...
+import sys
+import os.path
+from os import error as os_error
+
+from smbclient._os import remove, rmdir, listdir, stat, makedirs, readlink, symlink, scandir, copyfile_server_side
+from smbclient.path import islink, isdir
+from smbprotocol.exceptions import SMBOSError
+
+
+class Error(EnvironmentError):
```
jborean93: This seems quite generic, can we use one of the existing SMB Exceptions we have in `smbprotocol.exceptions`?

tests/test_smbclient_shutil.py
```diff
@@ -0,0 +1,142 @@
...
+def test_rmtree_non_existing_ignore_errors(smb_share):
+    rmtree("%s\\dir2" % smb_share, ignore_errors=True)
+
+
+def test_copy(smb_share):
+    mkdir("%s\\dir2" % smb_share)
+    with open_file("%s\\file1" % smb_share, mode='w') as fd:
+        fd.write(u"content")
+
+    sleep(0.1)
```
jborean93: Why the sleep here?

tests/test_smbclient_shutil.py
```diff
@@ -0,0 +1,142 @@
...
+
+    assert src_stat.st_size == dst_stat.st_size
+
+
+def test_copy_raises_when_source_and_target_identical(smb_share):
+    mkdir("%s\\dir2" % smb_share)
+    with open_file("%s\\file1" % smb_share, mode='w') as fd:
+        fd.write(u"content")
+
+    sleep(0.1)
```
jborean93: Sleep again

tests/test_smbclient_shutil.py
```diff
@@ -0,0 +1,142 @@
...
+    if (sys.version_info > (3, 0)):
```
jborean93: Why the version difference?

tests/test_smbclient_shutil.py
```diff
@@ -0,0 +1,142 @@
...
+    with context:
+        copy("%s\\file1" % smb_share, "%s\\file1" % smb_share)
+
+
+def test_copy_with_dir_as_target(smb_share):
+    mkdir("%s\\dir2" % smb_share)
+    with open_file("%s\\file1" % smb_share, mode='w') as fd:
+        fd.write(u"content")
+
+    sleep(0.1)
```
jborean93: Sleep

tests/test_smbclient_shutil.py
```diff
@@ -0,0 +1,142 @@
...
+    mkdir("%s\\dir2" % smb_share)
+    mkdir("%s\\dir2\\dir3" % smb_share)
+
+    with open_file("%s\\dir2\\dir3\\file1" % smb_share, mode='w') as fd:
+        fd.write(u"content")
+
+    with open_file("%s\\dir2\\file2" % smb_share, mode='w') as fd:
+        fd.write(u"content")
+
+    sleep(0.01)
```
jborean93: Sleep

tests/test_smbclient_shutil.py
```diff
@@ -0,0 +1,142 @@
...
+    with open_file("%s\\dir2\\dir3\\file1" % smb_share, mode='w') as fd:
+        fd.write(u"content")
+
+    with open_file("%s\\dir2\\file2" % smb_share, mode='w') as fd:
+        fd.write(u"content")
+
+    def ignore(src, names):
+        return [name for name in names if name == "file2"]
+
+    sleep(0.01)
```
jborean93: Sleep

smbclient/_os.py
jborean93: Ok I did some testing myself and can confirm those limits are still in place for SMB2. You can actually see the limit documented on the raw IOCTL docs for this command at https://docs.microsoft.com/en-us/windows/win32/devnotes/ioctl-copychunk.

The `Length` field states

> The number of bytes of data in the chunk to be copied. Must be greater than zero and less than or equal to 1 MB. Length * ChunkCount must be less than or equal to 16 MB.

So this confirms the max size of `1024 * 1024` for a chunk size and 16 chunks per request. Sorry for doubting you here :)

smbclient/_os.py
```diff
@@ -1016,3 +1017,93 @@ def stat(self, follow_symlinks=True):
...
+        val_resp = SMB2SrvRequestResumeKey()
+        val_resp.unpack(transaction_src.results[0])
+
+        chunks = _get_srv_copy_chunks(transaction_src)
+
+        with SMBRawIO(norm_dst, mode='x', desired_access=FilePipePrinterAccessMask.GENERIC_WRITE,
+                      share_access='r',
+                      create_options=CreateOptions.FILE_NON_DIRECTORY_FILE, **kwargs) as raw_dst:
+
+            with SMBFileTransaction(raw_dst) as transaction_dst:
```
jborean93: This line should be swapped with the for loop. Right now it will send a compound with multiple IOCTL requests which fails as soon as you go over 16 due to the `MAX_COPY_CHUNK_COUNT`. Swapping it means each request only contains up to 16MB of copy chunks hitting the max limit.

Also because of this you will have to move the transaction check below to the proper indentation.

smbclient/shutil.py
martinhoefling: non server side copy code removed for now

smbclient/shutil.py
martinhoefling: removed

smbclient/shutil.py
martinhoefling: Well, copytree under the hood uses copy2. I named it like that to make clear that it's not the same as shutil.copytree

tests/test_smbclient_shutil.py
martinhoefling: That's a pytest bug. On python2, pytest cannot properly handle unicode error messages. I add a comment with a reference to the bug.

smbclient/shutil.py
jborean93: Hmm I see what you mean. I still think we should have it called copytree and just make it explicit in the docs that it just doesn’t copy the file metadata for now. We can always add that in a future release once we have a good and working copy2 implementation.

The reason why I think we should have this but not necessarily a copy2 for now is that it is better to have partial functionality whereas copy2’s only extra feature over just copy is something that I’m not 100% sure how to implement right now.

tests/test_smbclient_shutil.py
jborean93: That shouldn’t be an issue, the `smbprotocol._text` has a `to_native` function which can be used on Python 2 and 3 to create a native string from a Unicode string. Technically it’s a no-op on py3 but on py2 it should overcome the limitation of non-ASCII chars in an exception string.

You should just be able to do `match=re.escape(to_native(expected))` but best to test it out first.

martinhoefling: With that change, I obtain the same error as before:
```
        expected = 'are the same file'
        with pytest.raises(ValueError, match=re.escape(to_native(expected))):
>           copy("%s\\file1" % smb_share, "%s\\file1" % smb_share)
E           UnicodeEncodeError: 'ascii' codec can't encode character u'\xfd' in position 30: ordinal not in range(128)
```

smbclient/shutil.py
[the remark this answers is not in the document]
martinhoefling: I have renamed it and removed all the functionality to copy between local and shares or between shares for now.

smbclient/_os.py
martinhoefling: done

smbclient/_os.py
martinhoefling: done

smbclient/shutil.py
martinhoefling: removed

smbclient/shutil.py
martinhoefling: ok, renamed to copytree and added note in the docstring

tests/test_smbclient_shutil.py
martinhoefling: removed

tests/test_smbclient_shutil.py
martinhoefling: I was not able to resolve this for now. More ideas? I think the corresponding issue is: https://github.com/pytest-dev/pytest/issues/5478

tests/test_smbclient_shutil.py
martinhoefling: removed

tests/test_smbclient_shutil.py
martinhoefling: removed

smbclient/_os.py
martinhoefling: swapped :-)

tests/test_smbclient_shutil.py
jborean93: Make sure you do to_native when you raise the actual exception itself. Also it’s best to make sure your expectation is a Unicode string first by putting the u prefix before it.

If that doesn’t work we can leave it for now.

martinhoefling:
@jborean93 , I hope I have addressed most issues. Two are still open I think:
- SMBRawIO usage: Maybe you can take a look, why the suggestions didn't work out.
- pytest bug: Same thing. Maybe we just leave it like that and omit checking the message on python 2.7?

smbclient/_os.py
martinhoefling: No worries. I have extracted it as constants and have added a proper description.

tests/test_smbclient_shutil.py
martinhoefling: I tried
```
    if _samefile(src, dst, **kwargs):
        raise ValueError(to_native(u"`%s` and `%s` are the same file") % (src, dst))
```
Same error.

jborean93: No worries, I'll have a look at it later and will just keep it there if I can't fix it. Thanks for trying though.

jborean93:
I'll have a look at both points once easy-api has been merged into the master branch before finalising the new release as well.

martinhoefling:
@jborean93 thx, for all your comments and help. Lets get that one released and I would then work on copy2.

jborean93:
@martinhoefling about those 2 remaining issues, here is the fix for the unicode chars in the exception https://github.com/jborean93/smbprotocol/commit/dd1b5de012144aacbaef6dcc7df14c7467c37333.

It wasn't an issue in pytest but rather how Python 2 handles native string compared to 3. In 2, a native string is actually a byte string and the default behaviour of converting a unicode string to byte is to encode it using the `ascii` encoding. This causes issues in cases where the unicode string contains non-ascii chars and then re-raises another ValueError for that. The way to get past that is to ensure you already pass in a native string (byte in 2, unicode in 3) so it doesn't try and encode it using ascii. The `to_native` method in smbprotocol handles all this for you and defaults to `utf-8` which is lot more sane in today's environment.

As for the SMBRawIO to open_file this was fixed in https://github.com/jborean93/smbprotocol/commit/39d82b7976292c19b9d8bfcbef7dd04125b97c43. Basically the SMBFileTransaction expects a SMBRawIO object and when you use open_file that can be encapsulated in a few different layers. In this case I am making sure I open the file in binary mode (`mode='*b'`) and I cna then access the raw IO object with the `raw` property. I can even take this a step futher and set `buffering=0` to `open_file` and that returns the relevant SMBRawIO object for the file type specified.

martinhoefling:
Thx!

Issue #12, opened 2019-10-21
Connection.receive does not break if wait=False


kvv81:
https://github.com/jborean93/smbprotocol/blob/9c06fc5ce3eab3ce2fa8abcc55c5323baeecdc89/smbprotocol/connection.py#L1024

If wait=False is passed as the parameter then this condition is always False so receive() function makes infinite loop (if no timeout, or exited after timeout otherwise).
I think that correct condition is as below:

`if status is not None and (not wait or status != NtStatus.STATUS_PENDING):`

jborean93:
It's been a while since I've last delved into that code but I think the current behaviour is correct. The `wait` kwarg is designed to continue waiting after receiving a `STATUS_PENDING` response from the server. If that response is received and `wait=True` is set then it will continue to wait until the request has been processed and the final response has been received. The `timeout` parameter is designed to control how long it waits for, if you set it to 0 then it should do what you are looking for.

kvv81:
I intend to wait for STATUS_PENDING reply packet and return it (without waiting for final STATUS_SUCCESS); still we have to receive SOME reply, I don't want to use any timeout value.
I think 'wait' parameter was designed exactly for that purpose, no?

In case if wait=True + timeout=0 we will have 'try to receive but exit immediately if reply is not received yet' logic which is NOT what I need (in this case we can exit without any reply packet).

Please comment what do we get if wait=False is passed - I still think this is not useful use-case at all.

jborean93:
@kvv81, sorry for the delay on this, I've just opened a PR which I believe solves your issue here. When looking at the [new receive code](https://github.com/jborean93/smbprotocol/blob/42a67e7ac29ef843becbb192bfb0581948918108/smbprotocol/connection.py#L984) it is designed to do the following:

* By default (`wait=True, timeout=None`) it will wait until a non-pending result is received indefinitely
    * If a pending result is received it will continue to wait until the non-pending response arrives
    * This is how it works today
* If you set `wait=False` it will return if a pending result or the final result is received from the server
    * I believe this is the issue you are mentioning here and this new behaviour will allow you to wait until just the pending result arrives
* If you set `timeout`, the whole receive operation will continue to wait until either the response has arrived or the timeout has been reached
    * If the timeout has been reached then an exception is raised
    * The pending vs non-pending is still controlled by the `wait` kwarg
    * If you receive a pending result and `wait=True`, it does not reset the timer. A timeout is calculated based on when receive was first entered to now

[...]

Issue #5, opened 2018-11-01
Interface is overly complex: couple of file operations takes 120 lines


sovetov:
Hi,

While the project is extensible and "correct" from OOP point of view, it lacks sensible defaults and shortcuts. In the examples in this repo, it took 120 lines to open, read and delete a file. This makes the project unusable... 

It would be cool to get a file this way, for example:
```py
with Session.connect('server', 12345, 'Administrator', password) as session:
    with session.open_file('C$', '/path/to/file') as f:
        ...
```

Having only two visible entities, session and opened file, would be enough for most users. `Connection` and `TreeConnect` may be hidden. Security stuff needs more shurtcuts as it's impossible for average user to quickly come up with correct ones.

jp495:
On a related note -- would you mind walking me through the pseudocode required to, say, retrieve the `FileName`s and `LastWriteTime`s of all files within a directory in a share?

I'm at the point where I'm connecting to the Tree just fine but running into issues querying from there. I'm new to SMB, and as well-documented as `smbprotocol` is, I'm having a hard time knowing where to start or what direction to head in.

jborean93:
@jp495 I really do wish there was an easier way to do this but I still haven't had time to come back and revisit this. It's still definitely on my radar to fix up and I'm hoping to get back into it at some point.

For your question, I created a gist https://gist.github.com/jborean93/a3cb93fa6237012ebf587e1fbe8fc903 that will list the contents of a directory over an SMB share. These are the sort of things I wished to have a better layer over the top of smbprotocol to help abstract out the complexities of SMB for end users.

jp495:
@jborean93 no worries! A higher level API is an easy thing to ask for when I'm not the one doing the work. I appreciate all the work you've already put into the `smbprotocol`.

And thank you for that gist. It's a huge help.

jborean93:
Just an FYI, I've started work on this and am slowly making progress. You can see the latest status by looking at the [easy-api](https://github.com/jborean93/smbprotocol/tree/easy-api) branch. Ultimately I'm trying to replicate the builtin `os` and `shutil` modules in Python.

jborean93:
I've finally gotten around to it and have finalised the PR to add the easy-api changes to smbprotocol. If you are interested in trying this out and need any help let me know and I'll walk you through some of it. The `smbclient` package should simplify all these operations in a major way and provide a nice an easy way to complete common file operations.

[...]

PR #17, opened 2019-11-22
Add smbclient for an easy high level API to manage SMB shares

jborean93:
This PR adds the `smbclient` package to `smbprotocol` which is designed to be an easy to use API that sits on top of `smbprotocol`. It tries to be a like for like copy of the `os`, `os.path`, and `shutil` interfaces so that people familiar with the builtin file operation functions in Python should be easily able to use `smbclient`.

For example, to open a file and read the data, it's as simple as:

```
import smbclient

with smbclient.open_file(r"\\server\share\file.txt", username="user", password="pass") as fd:
    data = fd.read()
```

With Kerberos auth you can even rely on implicit authentication and don't have to supply credentials on the first connection.

Other fixes included with this PR are;

* Officially dropped Python 3.4 support as it is EOL, a previous PR also dropped Python 2.6
* Moves a few classes around to better suite their proper purpose this should still be backwards compatible as the old locations still import them. Some of the moves are;
    * `smbprotocol.query_info` is not in `smbprotocol.file_info`, `query_info` imports all of `file_info` but this will go away in a future release
    * `smbprotocol.connection.Dialects` is now in `smbprotocol.Dialects`
    * `smbprotocol.connection.Commands` is now in `smbprotocol.Commands`
    * `smbprotocol.connection.NtStatus` is now in `smbprotocol.exceptions.NtStatus`
    * `smbprotocol.open.FileAttributes` is now in `smbprotocol.file_info.FileAttributes`
    * `smbprotocol.open.FileInformationClass` is now in `smbprotocol.file_info.FileInformationClass`
* Added more file information classes for the SMB2 Query and Set Info requests
* Automatically resolve symlinks encountered in paths, whereas previous behaviour was to fire an exception

Fixes https://github.com/jborean93/smbprotocol/issues/5
Fixes https://github.com/jborean93/smbprotocol/issues/12

Thanks to @martinhoefling for helping implement some of the `shutil` functions that are included in this PR.


jborean93:
Beta1 of this change is available through https://pypi.org/project/smbprotocol/1.0.0b1/ `pip install smbprotocol==1.0.0b1`. If all goes well then the final will be out at the end of this week.

PR #34, opened 2020-03-22
Feature/add full size stat

iddohau:
Hi,
I add support to FileFsFullSizeInformation via the stat function.
I have tested it on windows 10 share and on the docker image in the README.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ except ImportError:
 
 setup(
     name='smbprotocol',
-    version='1.0.2.dev0',
+    version='1.0.3',
     packages=['smbclient', 'smbprotocol'],
     install_requires=[
         'cryptography>=2.0',
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ except ImportError:
 
 setup(
     name='smbprotocol',
-    version='1.0.3',
+    version='1.0.2.dev0',
     packages=['smbclient', 'smbprotocol'],
     install_requires=[
         'cryptography>=2.0',
--- a/smbclient/_os.py
+++ b/smbclient/_os.py
@@ -113,7 +113,7 @@ SMBStatResult = collections.namedtuple('SMBStatResult', [
 SMBFullSizeResult = collections.namedtuple('SMBFullSizeResult', [
     'total_size',
     'caller_available_size',
-    'actual_available_size',
+    'actual_available_size'
 ])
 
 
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -16,7 +16,7 @@
 
 setup(
     name='smbprotocol',
-    version='1.0.2.dev0',
+    version='1.0.3',
```
jborean93: Best to leave this as is, technically the next release will now be `1.1.0` as this is a new feature but I deal with that when packaging a new version.

smbprotocol/file_info.py
```diff
@@ -877,3 +877,23 @@ def __init__(self):
...
+    [MS-FSCC] 2.5.4 FileFsFullSizeInformation
+    https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/63768db7-9012-4209-8cca-00781e7322f5
+    """
+
+    INFO_TYPE = InfoType.SMB2_0_INFO_FILESYSTEM
+    INFO_CLASS = FileSystemInformationClass.FILE_FS_FULL_SIZE_INFORMATION
+
+    def __init__(self):
+        self.fields = OrderedDict([
+            ('total_allocation_units', IntField(size=8)),
```
jborean93: By default an IntField is unsigned but according to the docs for FileFsFullSizeInformation the first 3 fields are a signed integer (Int64). You just need to add `unsigned=False` to the kwargs of IntField for the first 3 fields. The `sectors_per_unit` and `bytes_per_sectore` are documented as unsigned though so those as fine.

smbprotocol/file_info.py
```diff
@@ -877,3 +877,23 @@ def __init__(self):
             )),
         ])
         super(FileFsVolumeInformation, self).__init__()
+
+
+class FileFsFullSizeInformation(Structure):
```
jborean93: I've tried to keep these in the order they are documented in the relevant MS docs, this should go before `FileFsObjectIdInformation`.

smbclient/_os.py
```diff
@@ -597,6 +609,11 @@ def stat(path, follow_symlinks=True, **kwargs):
         st_chgtime_ns=chgtime_ns,
         st_file_attributes=file_attributes.get_value(),
         st_reparse_tag=reparse_tag,
+        st_full_size_total_allocation_units=full_size['total_allocation_units'].get_value(),
```
jborean93: I don't think having this as part of stat is a good idea, this information is info on the file system and not the file specified. While we do get the FsVolumeInformation that is purely because Python's `stat` returns the the volume ID for a file so it can compare whether 2 files are the same.

The trouble is I don't know how we want to expose this information. Do you have any ideas how you think it should look?

iddohau: Maybe I will create new function `get_full_size_information` that will return this information?

jborean93: Let's just do that for now and get your contribution in, I don't want to hold this up because I haven't decided how to expose it yet. If I decide to move it around in another PR before the next release I'll be sure to let you know in the PR that does that.

iddohau: Ok, Thanks

jborean93:
Sorry for screwing it up, got there in the end.

PR #39, opened 2020-05-01
Feature/connection cache

iddohau:
This feature will able to maintain a connection cache that isn't global.
This will give you more usage options as:
- Maintain multiple connection to the same share
- Connecting to the same share with different credentials without the need to logout
- The user can deiced how manage the connections

### DIFF
--- a/smbclient/_pool.py
+++ b/smbclient/_pool.py
@@ -19,28 +19,30 @@ from smbprotocol.tree import (
     TreeConnect,
 )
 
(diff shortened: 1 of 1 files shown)


jborean93:
I'm not ignoring this PR sorry, I've just got a few other things I need to deal with first before I review your 2 PRs. Sorry for the delay.

jborean93:
Really sorry for the delay here, changes look good to me.

PR #45, opened 2020-08-05
Migrate CI across to Azure pipelines

jborean93:


adiroiban:
Just a comment. Why not use GitHub Actions? Is the same infrastructure as Azure Pipelines, but provides better integration with GitHub and you don't need extra Azure Devops account.

jborean93:
> Why not use GitHub Actions? 

1. I have an existing framework for Python libraries that use Azure DevOps that already covers things like test/coverage reporting, wheel artifact creating, auto pypi releases on tagging
2. It's another tool that I need to spend time to learn whereas I know more about Azure DevOps right now
3. Having more integration into GitHub doesn't really give me any benefits that I know off

As for the extra Azure DevOps account that's not really an issue, I already have a Microsoft account so it's no big deal. Maybe it is better but for now I have a working solution with Azure DevOps and prefer not to reimplement it all :)

Issue #26, opened 2020-01-02
Add support for DFS shares


Fen0l:
Hello,

I am getting this error when I try to used it with my File Server.  On the Microsoft doc it seems related to STATUS_PATH_NOT_COVERED. Im trying with the full DFS share name but impossible. Is smbprotocol support DFS ? Or any idea of what should I used ?

`smbprotocol.exceptions.SMBResponseException: Received unexpected status from the server: (3221226071) UNKNOWN_ENUM: 0xc0000257`

Impossible to list the root of any share and any directories from share.

Thank for your hints :)

jborean93:
Unfortunately smbprotocol does not yet support DFS, it may work in some situations but things like referrals of when the DFS path is for a folder/file on another server it will fail. It's something I do want to add at some point in the future but this is not a trivial task. Just looking at https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dfsc/1ff7d611-ba19-49fb-95f4-7c5358b86834 shows you how many branching steps there are.

I will edit the issue title to better reflect this as a request to support DFS.

exislow:
DFS support would be great. Any chance for an implementation?

jborean93:
Not anytime soon unless someone is willing to do it sorry. I’ve got 3 other side projects I’ve started and want to complete before even looking at this.

exislow:
Too bad. Do you maybe know any Python alternative, which can handle DFS smb shares?

jborean93:
The only other python library I know that does SMB is [pysmb](https://github.com/miketeo/pysmb) but that does not support it.

DFS support is quite complex and not easy to implement. There’s a reason why it doesn’t work in a lot of 3rd party libraries. You can try and resolve one of the time servers behind your DFS name then you can access it like any normal SMB server.

exislow:
Okay, I understand.

I tried this but without any success somehow. How would you revolve a time server in DFS? For example:

- MyDFS URL is: big.corp
- Share Name: SHAREFS
- Path to access: MyFolder
=> \\big.corp\SHAREFS\MyFolder

How do I a resolve the time server to MyFolder and how do I access it later? I am happy about every hint!

jborean93:
[...]

PR #49, opened 2020-09-16
Add Experimental DFS Support

jborean93:
This is an extension of @iddohau work https://github.com/jborean93/smbprotocol/pull/37 for adding DFS support to `smbclient`. There was quite a bit more work required to get it running with both a standalone and domain based DFS server. I've also tried to set up a caching mechanism but it still needs some more testing to make sure it is ok.

I need to set up a Vagrantfile that builds a custom environment to test out things like domain based DFS setups as I cannot do that normally in CI.

Fixes https://github.com/jborean93/smbprotocol/issues/26
Supersedes https://github.com/jborean93/smbprotocol/pull/37


jborean93:
The functionality added here should work in the basic use cases but will need more real world usage to find any further bugs. I'm merging it in for the next release and is set as experimental in the README.

PR #57, opened 2020-11-19
Consistent connection_key naming

JensVanhooydonck:
The naming of the connection_key isn't consitent between register_session and delete_session


jborean93 (approved):

jborean93:
Thanks for the pickup.

PR #58, opened 2020-11-20
Allow passing kwargs to spnego.client from high-level API

samuelwcm:
I have added in a way to pass extra kwargs to `spnego.client`. The primary motivation behind doing this was to be able to disable NTLM fallback authentication while using the high-level API, but I think the low-level API doesn't have support for this either, so this PR will add this functionality in both places.

I have marked this as a WIP for now as I haven't written any tests, but I wanted to get it on your radar.

### DIFF
--- a/smbclient/_pool.py
+++ b/smbclient/_pool.py
@@ -251,7 +251,6 @@ def get_smb_tree(path, username=None, password=None, port=445, encrypt=None, con
         'encrypt': encrypt,
         'connection_timeout': connection_timeout,
         'connection_cache': connection_cache,
-        'session_config': session_config,
     }
 
     # Normalise and check that the path contains at least 2 components, \\server\share
--- a/smbclient/_pool.py
+++ b/smbclient/_pool.py
@@ -83,19 +83,20 @@ class ClientConfig(object):
         skip_dfs (bool): Whether to skip using any DFS referral checks and treat any path as a normal path. This is
             only useful if there are problems with the DFS resolver or you wish to avoid the extra round trip(s) the
             resolver requires.
-        session_config (Optional[dict]): Extra kwargs to pass to the Session.
+        auth_protocol (str): The protocol to use for authentication. Possible values are 'negotiate', 'ntlm' or
+            'kerberos'. Defaults to 'negotiate'.
     """
 
     def __init__(self, client_guid=None, username=None, password=None, domain_controller=None, skip_dfs=False,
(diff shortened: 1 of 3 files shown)


jborean93:
I'm not comfortable with exposing the mechanisms of `spnego` to a user of this library as it locks me into using `pyspnego` or risk making a breaking change to end users. The better approach to this is to expose the functionality you want (kerberos only) as an option and the session setup code translate that into whatever is needed when calling `spnego.client()`.

In this case I would recommend adding a kwargs that can select the protocol (`negotiate` (default), `kerberos`, or `ntlm`) and then just passing it through to `spnego.client(..., protocol=auth_procol)`.

samuelwcm:
Thanks for your comments, that makes sense to me. I'll work on implementing your suggested approach.

BTW, what level of testing (unit tests or otherwise) would you suggest I do to call this PR complete? So far I have spun up an integration environment using the Vagrantfile in the repo, and disabled NTLM on the Server 2019 box using a GPO. Would you like me to look at scripting up an environment like this with Ansible or would you be prepared to take my word for it that it works?

jborean93:
In this case I don't think we can really test it out in CI as Kerberos is a massive pain to set up. I do have a collection of integration tests at https://github.com/jborean93/smbprotocol/blob/master/tests/integration/templates/test_integration.py.tmpl. You could have a test with `auth_protocol='kerberos'` to make sure it doesn't fail. You could also have a test with an IP address that checks for a failure.

The issue is these tests aren't too easy to set up and run if you don't know much about Vagrant and Ansible. Happy to help out where I can.

samuelwcm:
I've implemented the suggested change, I'll look at writing some tests later. In the mean time I've slightly fleshed out the docs for the Vagrant environment. Would you like me to raise a separate PR for these or shall I include them here? https://github.com/samuelwcm/smbprotocol/blob/docs/tests/integration/README.md

jborean93:
Separate PR would be great. Keeps things self contained.

jborean93:
Thansk @samuelwcm for the changes. I've made a slight change where `ClientConfig()` and `register_session()` accept just `auth_protocol` rather than blanket kwargs. This does mean future args to `Session` would need to be exposed like this but I prefer to strictly control what we expose rather than exposing those internals to the caller.

samuelwcm:
Hey, thanks for merging this! Apologies for not getting round to adding tests for this, I moved house just after I submitted the PR and never found the time :(

Issue #64, opened 2020-12-30
What is the maximum data write limit?


ashar0706:
I am trying to download a very large file from a client to the remote server that I am creating using the smbclient, I break down my file into chunks, I perform the following steps:

1. Open an empty file in the remote share server
2. truncate the length of the destination file to source file length 
3. Now I want to write the chunk to their respective offset however I am unable to perform this

```
smbclient.truncate(path=destination_file_path, length=file_size)

with smbclient.open_file(path=destination_file_path, mode="ab", share_access="rw") as fd:
    fd.seek(chunk_start)
    fd.write(chunk_data_in_bytes)

```

I am able to write at most 8MBs of a data per chunk at a specified offset, when I try to write for more than 8Mb I get the following error

```
Traceback (most recent call last):
  File "/var/task/smbclient/_io.py", line 505, in write
    transaction += self.fd.write(b, offset=self._offset, send=False)
  File "/var/task/smbprotocol/open.py", line 1345, in write
    raise SMBException("The requested write length %d is greater than "
smbprotocol.exceptions.SMBException: The requested write length 20971520 is greater than the maximum negotiated write size 8388608
```

Wanted to confirm if 8MB is the write limit?

jborean93:
There is no real default limit, the size is part of the negotiation process and returned by the server https://github.com/jborean93/smbprotocol/blob/00c1828a12c245e839eae1de85098b2750effbfe/smbprotocol/connection.py#L456-L458. You can always guarantee the limit to be 64KiB or higher. If you are wanting to get the max read/write size you can open the file in raw module like so

```python
import smbclient
import shutil

buffer = b'blah'

# buffering=0 means raw mode and there's no buffering done by Python (all on the OS level)
with open('local_path', mode='rb', buffering=0) as src_fd, smbclient.open_file(dest_path, mode='ab', share_access='rw', buffering=0) as dst_fd:
    # Use max_read_size if you want to read something from SMB
    buffer_size = dst_fd.fd.connection.max_write_size
    shutil.copyfileobj(src_fd, dst_fd, buffer_size)
```

I believe the reason why I didn't use the max_*_size attributes for `smbclient.shutil.copy()` is because while that might be the max size available the client may not have enough credits available to read/write something of that size. I've not yet fully understand the credit system used by SMB so for a stable and sane default I kept it to the max SMB packet size which should always be available.

mas-dse-juremigi:
I am also experiencing this issue:

[...]

PR #65, opened 2021-01-16
Limit write to max_write_size and provide writeall() to work around it.

dHannasch:
This is one possible way to close https://github.com/jborean93/smbprotocol/issues/64.

https://github.com/jborean93/smbprotocol/issues/64#issuecomment-758380434
> the reason why I didn't use the max_*_size attributes for `smbclient.shutil.copy()` is because while that might be the max size available the client may not have enough credits available to read/write something of that size. I've not yet fully understand the credit system used by SMB so for a stable and sane default I kept it to the max SMB packet size which should always be available.

I don't understand the credit system at all, so I don't know what other complications there might be.

Curiously, the underlying Python `io` library provides for `readall()` but not `writeall()`.

In theory, every `write()` ultimately returns an actual number of bytes written which might differ from the number of bytes you *tried* to write. But in general, there seems to be very little support for "No, really, write all of these bytes." Even `shutil`doesn't bother to check whether all of the bytes were actually written.

https://github.com/python/cpython/blob/master/Lib/shutil.py#L202-L208
```
    fsrc_read = fsrc.read
    fdst_write = fdst.write
    while True:
        buf = fsrc_read(length)
        if not buf:
            break
        fdst_write(buf)
```
If `fdst.write` fails to write all the bytes, `copyfileobj(fsrc, fdst, length=0)` will just go blithely along without even noticing.

And on reflection, maybe that's not surprising. In the canonical case of writing to a local disk, there's no point in something like `writeall()`. If you're out of disk space, then you're out of disk space; trying again isn't gonna get you anywhere.

Really, the Samba server behavior sounds weirder the longer I think about it. There's some "maximum negotiated write size 8388608" (= 2<sup>23</sup>).

https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/bcd6d594-017b-47fc-8742-b7d847791783
> MaxWriteSize is set to the maximum size, in bytes, of the Length in an SMB2 Write Request (2.2.21) that the server will accept on the transport that established this connection. This value SHOULD<237> be greater than or equal to 65536.
https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/a64e55aa-1152-48e4-8206-edd96444e7f7#Appendix_A_237
> | Windows version\Connection.Dialect | MaxWriteSize |
> | --- | ----------- |
> | Windows 7\Windows Server 2008 R2 | 1048576 |
> | Windows 8 without [MSKB-2934016]\Windows Server 2012 without [MSKB-2934016] | 1048576 |
> | All other SMB2 servers | 8388608 |

This is not a limit on an individual payload. *That* is sensibly limited to 65536. So all write requests are broken up into payloads of 65536 bytes.

This is not a limit on file size on the accessed file system, nor a limit on disk space. *Those* will naturally come up in due course when they apply (and of course might potentially be smaller than 8388608). This is an entirely separate, hardcoded limit *packets sent*, but on the *write request* itself.

As far as I can tell, the Samba server is completely 100% happy to receive any number of back-to-back write requests (until you hit the limit on file size or space on the underlying file system). They just want to receive multiple separate requests, for...no reason that the Microsoft docs there care to share. So maybe the reason the general base classes don't contemplate this is because this is an idiosyncrasy of Samba.

Testing this is a tricky question. We probably don't want to actually try to copy a 2<sup>23</sup> file in a test. But maybe we could add an additional optional parameter to `write()` and limit `write()` to the minimum of that and the actual MaxWriteSize. Similar idea to the (now-defunct?) `smbclient._os.CHUNK_SIZE`.


jborean93:
First off thanks for working on this and coming up with a PR, it's greatly appreciated. Also sorry for the block of text in my response, I'm hopefully going to be able to explain the current setup and we can go from there to try and find a good solution to this.

> I don't understand the credit system at all, so I don't know what other complications there might be.

I also can't claim to fully understand it but the general gist is that each SMB message (except the first negotiation message and cancel requests) must have a credit charge associated with it. The client must have credits to be able to send a message to the server and after the negotiation is complete it has 1 credit. Each message send will consume 1 credit but also request 1 credit to replace that so the client always has 1 credit available. The credit is based on the size of the request or the expected size of the response as per the [SMB2 protocol docs](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/18183100-026a-46e1-87a4-46013d534b9c?redirectedfrom=MSDN).

> CreditCharge = (max(SendPayloadSize, Expected ResponsePayloadSize) – 1) / 65536 + 1

A request can contain a recharge request for more than 1 credit if the client desires. Looking at [93 MS-SMB2](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/a64e55aa-1152-48e4-8206-edd96444e7f7#Appendix_A_93) it seems like Windows always requests 4 credits when it creates a session.

> <93> Section 3.2.4.1.2: Windows-based clients require a minimum of 4 credits.

The `smbprotocol` implementation is slightly different due to my rudimentary understanding of the credit system. As part of setting up the session it requests [256 credits](https://github.com/jborean93/smbprotocol/blob/cd5af9f81cb1a030e74f7235fecaca5667c262d3/smbprotocol/session.py#L280) credits for the client which in hindsight is probably a bit too much and needs to be reviewed. If the server grants this amount of credits, as denoted in the response, then a client could theorectically read/write up to 16MiB, or the max read/write size as per the negotiation whatever is smaller. I chose 256 due to my lack of understanding of the system and I've just never had the time to reinvestigate this problem so it's stuck so far.

The low level API is mostly agnostic to the credit charge, the only special casing is for the session setup that requests 256 and gets back what the server deems acceptable. If someone using this API wants to send a large file they should have enough credits by default but that isn't a guarantee. They even have the ability to request more credits with an [echo request](https://github.com/jborean93/smbprotocol/blob/cd5af9f81cb1a030e74f7235fecaca5667c262d3/smbprotocol/connection.py#L940).

So for a write request a client can write up to `min(max_write_size, available_credit_size)`. You can see what happens when they try to write `max_write_size` bytes but not have the available credits granted from the server https://gist.github.com/jborean93/255a104f69a791868e92ac5d37155963#gistcomment-2965564.

> smbprotocol.exceptions.SMBException: Request requires 128 credits but only 33 credits are available

The high level API shouldn't expose all this logic to the client and it should handle it all internally. It seems like for read requests I'm arbitrarily limiting the read to 64KiB unless the caller is opening the file with `buffering` being greater than 64KiB. This is probably not ideal as we could theorectically read something larger than this in 1 request but I'll talk about this a bit more.

For write requests it seems like we don't do any buffering at all and just write whatever the user specifies with causes the current problem if they are trying to write a really large block of data.

The other spanner in the works here is the buffering that the Python `io` classes do. The default is `-1` which means the in memory buffer is 64KiB so the underlying SMB2 write operation is not called until either

* More than 64KiB bytes have been written to the buffer
* The caller calls `flush()`
* The io stream is closed

I think I was under the assumption that if someone tries to write more than 64KiB then the internal io mechanism will write it in 64KiB chunks at a time but after testing that assumption is wrong leading to the problems in https://github.com/jborean93/smbprotocol/issues/64. The caller could even open the file with `buffering=0` to remove that buffer entirely and any calls to `write()` will be done then no matter the size of the data.

So what I think we should do here is have `write()` chunk up the data if it exceeds the max write size and just send a request for each chunk. If a user of the high level API wants to write data then they shouldn't have to worry about calling certain functions or specify extra arguments to do so. The `write()` function should handle all the complex protocol logic itself and just write the data somehow.

Now the question leads into how we should chunk the file. In my mind we should be trying to make this as efficient as possible and read/write as many bytes as possible but the client may not have enough credits to write up the the `max_write_size` specified by the server. I see the following possible solutions to this issue

* Always chunk at 64KiB, this is what we do for read requests and is guaranteed to always work
    * The downside with this approach is it's highly ineffecient as we are sending a brand new request per 64KiB compared to 8MiB that newer servers allow
* Use the `buffering` kwarg when opening the file to control the chunk size
    * `-1` will use 64KiB (or maybe the logic below)
    * `0` is raw and will write it as 1 chunk, behaviour today
    * `>=1` is the chunk size based on what the caller thinks is best
    * Allows the client very fine control over the behaviour but is quite low level
    * The `max_write_size` is not readily available to the client until they've opened the file
* Have logic to chunk dynamically based on the max write size/credits available
    * Follows the `min(max_write_size, available_credit_size)` logic
    * Theorectically allows us to send the maximum number of bytes the client it can
    * It might cause havok with multithreading but this library isn't guaranteed to be thread safe so I think we are fine there
    * We could even request credits ourselves to dynamically adjust the chunk size if the server grants more than we have available

I feel like the proper option here would be the last one and we should adjust our initial credit request during the session setup to 4 instead of 256 so we aren't consuming all of the server's credits. The biggest downside to this route is it complicates the code a lot more but I feel like it allows the internal logic to be as efficient as it needs to be without exposing all this complex details to the client. Basically if we are trying to write data I see the following SMB packets being exchanged:

* SMB Negotiate request
    * Responds with `max_write_size` based on the server configuration
* SMB Session setup request
    * Instead of 256 we request just 4 based on the documented behaviour for Windows
    * The server will either give us 1-4 credits based on its load
* SMB Tree connect request
* SMB Open request to file
* SMB Write request (for each chunk)
    * Data length is calculated as `min(len(data), max_write_size, ((available_credits - 1) * 65536))`
    * We subtract 1 from available_credits as we send a query request alongside our write, maybe this should only be sent at the end?
    * The credit request is `credits_consumed + ceiling(len(data_remaining) / 65536)`
    * The server should responds with the number of credits it has granted which could be 1 to the number requested
    * This loops until all the data has been sent
* SMB Close request for file
* SMB Tree disconnect
* SMB Session close
* SMB Connection close

There's one more additional complication to this logic, when we send the write request we send it as part of a compound request with a query for the `FileStandardInformation` so that will consume 1 credit in itself. This requires some more thought as I think we only need to send this right at the end. From what I can remember it's so we can properly set the end of file length internally so seek operations work properly.

What I need to do is capture some traffic between a Windows client and server when writing a large file to see how it acts in this situation. I'll probably do so today and comment here with those details.

jborean93:
I've just opened https://github.com/jborean93/smbprotocol/pull/69 which tries to implement what I'm thinking this should be. I still need to think more about `readinto()` and how to make reading a file in chunks more efficient. This is important for the copy scenario as we want to read a file as efficiently as possible but we don't want to call `readall()` and load all the bytes into Python's memory.

jborean93:
Thanks a lot for contributing, I've played around with this a lot and have found some inefficiencies in how the current code operated. They have been fixed with https://github.com/jborean93/smbprotocol/pull/69. Ultimately it changed the raw IO base to write as must data as it can and the `BufferedWriter` that wraps it will call it as many time as needed until all the data is written. No need for a `writeall()` function or even limit `write()` to a smaller amount of bytes.

It's also implemented the credit handling that I mentioned above so subsequent calls should be able to send more data if the server has granted those credits.

dHannasch:
I'm using the new `write()` in place of `writeall()` now in https://gitlab.com/dHannasch/samba-test/-/blob/master/test_samba_write.py#L39, and it's working great! Thank you very much!

(The weird complications in that file are because of working with a Samba server that dynamically dispatches to multiple servers based on the path. It seems that whatever first gets passed to `listdir()` determines which server steps up, though I haven't been able to figure out what's going on under the head, whether it's an nginx reverse proxy or what. I mention this because I think there's some hidden/forgotten memory that `smbclient.reset_connection_cache()` isn't resetting, because even after `reset_connection_cache()` it keeps trying to connect to the same server it connected to before and keeps getting `STATUS_PATH_NOT_COVERED`, but exiting the entire Python process and starting a new Python process makes it work. I haven't tried `delete_session()` against the crash, but it looks like `delete_session()` does basically the same thing so I'm not expecting that to work either. Right now, of course, I can't reproduce the crash at all, even on that particular server, never mind set up a test server to reproduce it. The crash comes and goes. I'll holler if I ever find a way to reliably reproduce the crash.)

jborean93:
Glad to hear that it works.

For the reset_connection_cache issue, I'm not sure what is happening there. That function essentially goes through all the connections in https://github.com/jborean93/smbprotocol/blob/master/smbclient/_pool.py#L50 and closes the connection. Subsequent connections should be creating a new socket to the server and subsequently a new connection/session. Maybe there is some DNS caching when it comes to connecting to the socket? Creating the socket is quite straightforward https://github.com/jborean93/smbprotocol/blob/92ed80ba4681edc7a7c40eb8fefc1c66a417f064/smbprotocol/transport.py#L56-L60.

PR #66, opened 2021-01-16
Remove smbclient._os.CHUNK_SIZE.

dHannasch:
smbclient._os.CHUNK_SIZE appears to be completely nonfunctional. It is never read anywhere in the codebase. I'm guessing smbclient._os.CHUNK_SIZE is a leftover from a previous implementation that used it.


jborean93:
Looks like it may have been used to control the server side chunk copying but you are right it's no longer used at all. Thanks for the pick up.

PR #67, opened 2021-01-16
Just add a couple of new tests without adding functions.

dHannasch:
Testing writing is actually a bit complicated, so this is split off where it can be examined individually. This just adds a couple of tests to write files. The idea is that a separate branch will provide a function that can be a drop-in replacement for what we're doing manually here.


jborean93:
Thanks for adding these tests, I've had to slightly tweak them based on recent changes but the bulk of the work was already there.

PR #68, opened 2021-01-16
writeall() as a separate function without changing write()

dHannasch:
This is built atop https://github.com/jborean93/smbprotocol/pull/67. Look there first. They're easier to read individually.

This might be a less invasive alternative to changing the behavior of `write()`.


jborean93:
Thanks for this but see https://github.com/jborean93/smbprotocol/pull/65#issuecomment-762587542 why I'm closing this.

PR #69, opened 2021-01-18
Make the read and writes more efficient

jborean93:
This PR expands on the raw read/write operations in `SMBRawIO` which is what is ultimately used by `smbclient.open_file()`. Before we had `write()` and `readinto()` naively just read and write the based on the buffer provided rather than only write what it can based on the server limits. The same code was also copied with `readall()` when it should be shared with `readinto()`.

This PR changes `write()` to write as many bytes as it can based on the server configured limits and the current credits available as well as request more credits so subsequent calls can fit more data into each request.

The `readinto()` method (used by `read(num)` is also the same in that it only reads as much data as it can in 1 request and also request more credits if need be. The `readall()` will continue to have the same behaviour but the code is just simplified to call `readinto()` internally so it also benefits from the credit optimisation.

Even if these methods only use 1 syscall at a time, the buffered readers/writers that encapsulate the raw stream handle all the buffering and logic to call them multiple times if need be.

What this means for reading and writing large amounts of data:

* Calling `write()` will now be able to write all the bytes specified, even if it exceeds the credit or max write size limit
    * For a fd opened with `buffering=0` then it may not write all the data at once but that's the expected behaviour
    * For buffered streams (the default) then it handles it as expected
    * Ultimately smbclient will now send as much data as possible given a request and try to make future calls fit even more data if it can
* Calling `read(num)` will be able to read as many bytes as possible
    * This may result in mutliple calls but it's based on whether you want all the bytes or based on a buffer size

Fixes https://github.com/jborean93/smbprotocol/issues/64


jborean93:
I need to check how the new code works with named pipes, there are some edge cases for pipes when dealing with read/writing the data as they don't have an offset of end of file.

jborean93:
Testing the named pipe stuff manually and I'm pretty happy how it works.

| Function | Behaviour |
|-|-|
| `read()` | Calls `SMB Read` continuously until the pipe is closed |
| `read(10)` | Calls `SMB Read` continuously until 10 bytes have been read |
| `read1()` | Calls `SMB Read` once until some data has returned |
| `read1(10)` | Calls `SMB Read` once until some data has returned, will return 10 bytes and store the rest in a buffer |

All these functions are blocking until data is read or the pipe has been closed. The `*1` functions will ensure it makes 1 SMB read request and return what was given (or the pipe was closed).

Issue #74, opened 2021-01-31
auth_protocol="kerberos" does not seem to work


grawity:
By default, the module can successfully use Kerberos whenever a TGT is present -- I'm able to open files without specifying any username or password.

However, if I specifically _request_ `auth_protocol="kerberos"`, it's no longer able to connect -- I receive "Major (589824): Invalid token was supplied, Minor (100001): Success".

I'm testing against Windows Server 2019, and this happens immediately after receiving the Negotiate Protocol Response from the server.

Full backtrace:

```
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/spnego/_context.py", line 88, in wrapper
    return func(*args, **kwargs)
  File "/usr/lib/python3.9/site-packages/spnego/gss.py", line 435, in step
    out_token = self._context.step(in_token)
  File "<decorator-gen-15>", line 2, in step
  File "/usr/lib/python3.9/site-packages/gssapi/_utils.py", line 155, in check_last_err
    return func(self, *args, **kwargs)
  File "<decorator-gen-5>", line 2, in step
  File "/usr/lib/python3.9/site-packages/gssapi/_utils.py", line 128, in catch_and_return_token
    return func(self, *args, **kwargs)
  File "/usr/lib/python3.9/site-packages/gssapi/sec_contexts.py", line 519, in step
    return self._initiator_step(token=token)
  File "/usr/lib/python3.9/site-packages/gssapi/sec_contexts.py", line 535, in _initiator_step
    res = rsec_contexts.init_sec_context(self._target_name, self._creds,
  File "gssapi/raw/sec_contexts.pyx", line 245, in gssapi.raw.sec_contexts.init_sec_context
gssapi.raw.exceptions.InvalidTokenError: Major (589824): Invalid token was supplied, Minor (100001): Success

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/smbprotocol/session.py", line 273, in connect
    out_token = context.step(in_token)
  File "/usr/lib/python3.9/site-packages/spnego/_context.py", line 91, in wrapper
    reraise(SpnegoError(base_error=native_err, context_msg=context))
  File "/usr/lib/python3.9/site-packages/spnego/_compat.py", line 60, in reraise
    raise exc
spnego.exceptions.InvalidTokenError: SpnegoError (9): Major (589824): Invalid token was supplied, Minor (100001): Success, Context: Processing security token

[...]

PR #76, opened 2021-02-01
Fix up auth logic for non-negotiate auth

jborean93:
The GSS Negotiation Token from the negotiation response is a [NegTokenInit2](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-spng/8e71cf53-e867-4b79-b5b5-38c92be3d472) token that contains a list of mechs the server supports and is parsed by the Negotiate process when selecting the optimistic protocol. When someone explicitly requests `ntlm` or `kerberos` auth we were still passing in this token but neither pure Kerberos or NTLM is able to parse the SPNEGO token structure causing a failure when generating the first token.

This PR blanks out that token when not using `negotiate` auth and fixes a logic bug that becomes present when using pure NTLM authentication not wrapped in SPNEGO.

Fixes https://github.com/jborean93/smbprotocol/issues/74

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Fixed up secure negotiation logic when connecting to older SMB dialects
 * Will attempt to perform secure negotiation even on older dialects that may not implement it properly
 * Added `ClientConfig` option `require_secure_negotiate` to globally turn off secure negotiation if the client wishes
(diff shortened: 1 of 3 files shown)


jborean93:
Confirmed fixed user's issue with https://github.com/jborean93/smbprotocol/issues/74#issuecomment-771151257.

Issue #78, opened 2021-02-05
Any way to check for or handle connection errors while manipulating files?


zuroh:
I have a scenario where i am writing files to a remote SMB share, however I would like to handle a case where the file writing stops and exception is raised when the connection between the two machines are interrupted for example. Currently the program just freezes and no exception is raised when this happens. I have also tried to execute the writing of file to SMB share in a separate thread and call that thread with a timeout but that doesn't seem to work either. 

Is there anyway to handle this issue?

I am connecting from a windows client to a linux server over local area network. 

```
import time
import smbclient
import threading
import _thread

class timeout():


    def __init__(self, time):
        self.time = time
        self.exit = False

    def __enter__(self):
        threading.Thread(target=self.callme).start()

    def callme(self):
        time.sleep(self.time)
        if self.exit == False:
            _thread.interrupt_main()  

    def __exit__(self, a, b, c):
        self.exit = True


with timeout(20):
    try:
        while True:
            time.sleep(0.3)
            try:
                smbclient.stat(
                    somefilehere)
            except Exception as e:
                print(e)
            print("done inner")
    except KeyboardInterrupt:
        print("done outer")
```

jborean93:
It should be raising the exception but it sounds like there is some bad logic in the connection handler that causes it to ignore and just perpetually wait in the background. This layer is quite complex as it uses a background thread to read from the socket and another thread to process those messages. Ultimately the bug needs to be found and fixed.

jborean93:
I've just opened https://github.com/jborean93/smbprotocol/pull/80 which exposes the `SMBConnectionClosed` that is raised on any connection errors. You should be able to catch this and retry for as long as you desire. It should also fix the freezes but it would be great if you could test it out and see if it works for you.

zuroh:
Thank you! I'll test it soon as I'm free and get back to you with the results.

zuroh:
Hi there, I did some testing with version 1.5.0

I found that if the connection is closed on the client's side, the exception is thrown almost instantly and it works as expected.

However if the connection is closed on the server side, then the exception takes around 12-15 seconds and very occasionally still hangs. 

Both give the same exception message
// SMB socket was closed, cannot send or receive any more data

Luckily for my use case i only needed to check the connection being closed on the client side. But out of curiosity is there a reason for the delay? Is there some sort of manual timeout value to manipulate?

jborean93:
[...]

Issue #56, opened 2020-11-17
Recv_thread: Unpack requires a buffer of 4 bytes


DragosFlorea:
Hi,
I've encountered a issue while using the library and I cannot figure it out why.
The error encountered is this:
![image](https://user-images.githubusercontent.com/19181854/99384918-d5cd9880-28d8-11eb-8e98-010123233216.png)

I've also checked the code and I found this. Could you describe a little bit why this is happening?

![image](https://user-images.githubusercontent.com/19181854/99384963-ed0c8600-28d8-11eb-8c07-459481429020.png)
Because of this warning the recv_thread is stopped.

The scenario is the following:
I have a grpc server that receives requests to download files from a share.
The download part is done via streaming of chunks of bytes.
The grpc server is configured to receive max 10 requests at a time and every requests creates his own thread, in which a smb connection is opened and a read -> stream process is starting.

The issue is visible when I try do download multiple large files, after a while the warning appear(notthing else) and all the requests for that share stops.
I've observed also that if I use only one grpc request, I didn't encountered the issue.

Thanks

adiroiban:
Hi @DragosFlorea , Thanks for reporting the issue.

I think that this warning should be updated to also log the stack trace like `log.warning('There be dragons.', exc_info=True)`

This should provide more info about the place where this is raised.

Are you able to patch your code and run it again to get a better stacktrace?

---------

Most probably this is due to some protocol features that are not supported (yet) by smbprotocol

Which SMB version do you use ?

Can you add more info about how to reproduce this error?

Cheers!

DragosFlorea:
Hi @adiroiban,
I can patch the code locally, with that property and come back with more log info.
I don't set the smb version, the code choose it automatically and I remember that it picks the last 3.1.1.

Regarding the reproducing part on your side, kind of difficult.
Multiple threads that connect to the same share, trying to download files locally, important is to have large files which keep the socket connection open for a long period of time. 

Thanks, I will come back with a more descriptive log.

DragosFlorea:
hi,
Until the warning occurs the threads looks like this (also i confirm that the smb version used in connection is SMB_3_1_1):
![image](https://user-images.githubusercontent.com/19181854/99415437-74212480-2900-11eb-90d8-8540848e6c0c.png)
Green are the threads created by the grpc server, which means there are aprox 10 files in processing
Red are the threads created by the smbprotocol library

And the warning is this:
[...]

PR #80, opened 2021-02-10
Fix TCP Receiver Hangs

jborean93:
The current mechanism used to read responses from the TCP socket is quite complex but summarised in a very simplified matters involves 2 different threads

* TCP thread
  * Constantly calls `socket.recv()` to read SMB responses and adds them to an internal queue once received
* Connection thread
  * Reads from the internal queue for any messages from the TCP thread
  * Unpacks each message and fires the request event saying it has been received

This whole mechanism does work but it's very brittle to any failures in any of those steps. If the socket is closed for whatever reason the connection thread will be notified and stop listening for more messages. The issue here is that the main thread waiting for the response is never notified of this failure causing it to hang indefinitely. There are probably some edge cases in the TCP recv code where it's not reading the full number of bytes requested causing a failure as seen in #56.

What this PR does is

* Removes the TCP thread, the connection thread will call recv() directly instead of using a queue
* Tries to harden the existing `socket.recv()` code to handle cases when the full length of bytes requests was not met
* Implements a timeout functionality to make up for the timeout offered by the internal queue
  * Needed to send the SMB Echo request every 10 minutes
* The connection thread will make sure any outstanding request wait events are set when it finishes
  * The current logic only did so on an exception and didn't handle a graceful shutdown from the server causing a hang if there were still outstanding requests
* The `SMBConnectionClosed` exception is raised when trying to send or receive anything and the socket has been closed
  * Related to the above point, if there were any outstanding requests then trying to receive that will raise `SMBConnectionClosed`
* `smbclient` will ignore any connections in the cache with a closed socket, it will create a new connection automatically.
* Closing (`.disconnect()`) a connection with a closed socket will no longer fail

One last thing I need to investigate is the code around closing the TCP socket. I've seen some sporadic messages printed to the console that seem to be exceptions but aren't treated as such. I believe it might be related to the `atexit` connection close operations for `smbclient` but I haven't fully tracked that down yet.

Fixes https://github.com/jborean93/smbprotocol/issues/56
Fixes https://github.com/jborean93/smbprotocol/issues/78

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## 1.5.0 - TBD
+
+* Added `smbprotocol.exceptions.SMBConnectionClosed` that is raised when trying to send or receive data on a connection that has been closed
+* Do not attempt to reuse any cached connections that have been closed in `smbclient`
+* Added a lock when writing to the socket, only 1 thread can write a message at a single point in time
+* Revamped the SMB receiver code to simplify the logic and make it more durable
+    * Removed the TCP recv thread for each connection, now each connection uses just 1 thread instead of 2
+    * Be more defensive when reading data from a socket to ensure we get all the data we require
+    * Handled server side FIN packets that close the connection unexpectedly, any requests waiting for a response will raise `SMBConnectionClosed`
+
+
 ## 1.4.0 - 2021-02-02
 
 * Fixed up secure negotiation logic when connecting to older SMB dialects
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -43,6 +43,8 @@ stages:
           python.version: 3.7
         Python38:
           python.version: 3.8
+        Python39:
+          python.version: 3.9
 
     steps:
     - task: UsePythonVersion@0
@@ -146,6 +148,12 @@ stages:
         Python38-x64:
           python.version: 3.8
           python.arch: x64
+        Python39-x86:
+          python.version: 3.9
+          python.arch: x86
+        Python39-x64:
+          python.version: 3.9
+          python.arch: x64
 
     steps:
     - task: UsePythonVersion@0
@@ -231,7 +239,7 @@ stages:
     steps:
     - task: UsePythonVersion@0
       inputs:
-        versionSpec: 3.8
+        versionSpec: 3.9
 
     - script: |
         python -m pip install twine wheel -c tests/constraints.txt
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ with open(abs_path('README.md'), mode='rb') as fd:
 
 setup(
     name='smbprotocol',
-    version='1.4.0',
+    version='1.5.0',
     packages=['smbclient', 'smbprotocol'],
     install_requires=[
         'cryptography>=2.0',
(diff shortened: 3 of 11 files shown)


jborean93:
I've tried to replicate the `sock.shutdown()` or `sock.close()` errors I saw sporadically in my testing but I cannot seem to get it to happen anymore. I've attempted to replicate it manually by trying all sorts of combinations of local sockets I fully control but the only time I see an error is when calling `sock.shutdown(socket.SHUT_RDWR)` multiple times. Because there's now a lock on the close method this should never occur so maybe I saw this error before I added the lock around this method.

Will add the tests and fix the existing ones but that was the only outstanding problem I had left to check.

jborean93:
Let me just say dealing with sockets at this level is quite painful. The differences in behaviour between the various platforms is annoying and trying to deal with all these niche cases is just crazy. I won't be surprised if something else pops up but for now the tests are passing and I'm fairly happy with the recent test additions that the main scenarios are covered.

adiroiban:
Hi. Thanks again for working on smbprotocol.

I have recently start seeing on my Windows tests the overlap of `smbprotocol.tranport.Tcp.close()` with `smbprotocol.tranport.Tcp.recv()` ...the issue is that `recv()` will only free the `Tcp._close_lock` after a read timeout.

I am trying to see how this can be fixed.

One idea is to lock only after the `select()` call.


```diff
def _recv(self, length, timeout):
    buffer = bytearray(length)
    offset = 0
    while offset < length:
        read_len = length - offset
        log.debug(f"Socket recv({read_len}) (total {length})")

        start_time = timeit.default_timer()

-        with self._close_lock:
-            if not self.connected:
-                # The socket was closed - need the no cover to avoid CI failing on race condition differences
-                return None, timeout  # pragma: no cover
-
-           read = select.select([self._sock], [], [], max(timeout, 1))[0]
-            timeout = timeout - (timeit.default_timer() - start_time)
-            if not read:
-                log.debug("Socket recv(%s) timed out")
-                raise TimeoutError()

+       read = select.select([self._sock], [], [], max(timeout, 1))[0]
+       timeout = timeout - (timeit.default_timer() - start_time)
+       if not read:
+            log.debug("Socket recv(%s) timed out")
+            raise TimeoutError()
+
+        with self._close_lock:
+            if not self.connected:
+                # The socket was closed - need the no cover to avoid CI failing on race condition differences
+                return None, timeout  # pragma: no cover
            try:
                b_data = self._sock.recv(read_len)
            except OSError as e:
                # Windows will raise this error if the socket has been shutdown, Linux return returns an empty byte
                # string so we just replicate that.
                if e.errno not in [errno.ESHUTDOWN, errno.ECONNRESET]:
                    # Avoid collecting coverage here to avoid CI failing due to race condition differences
                    raise  # pragma: no cover
                b_data = b""
```

adiroiban:
I have to do some more unit testing to see how Windows network api works.

On linux I see https://man7.org/linux/man-pages/man2/select.2.html


```
   Multithreaded applications
       If a file descriptor being monitored by select() is closed in
       another thread, the result is unspecified.  On some UNIX systems,
       select() unblocks and returns, with an indication that the file
       descriptor is ready (a subsequent I/O operation will likely fail
       with an error, unless another process reopens the file descriptor
       between the time select() returned and the I/O operation is
       performed).  On Linux (and some other systems), closing the file
       descriptor in another thread has no effect on select().  In
       summary, any application that relies on a particular behavior in
       this scenario must be considered buggy.
```

PR #82, opened 2021-03-07
An index cannot equal the length of the list.

dHannasch:
If it does, it's off the end.

### DIFF
--- a/smbclient/_pool.py
+++ b/smbclient/_pool.py
@@ -150,7 +150,7 @@ class ClientConfig(object):
         for referral in self._referral_cache:
             referral_path_components = [p for p in referral.dfs_path.split("\\") if p]
             for idx, referral_component in enumerate(referral_path_components):
-                if idx > len(path_components) or referral_component != path_components[idx]:
+                if idx >= len(path_components) or referral_component != path_components[idx]:
                     break
 
             else:


jborean93 (approved):

jborean93:
Thanks for the pickup here, are you able to share the scenario and error message you got when your encountered this error? From looking at the code it seems like it will fire if the DFS cache has an entry for `\\server\share\folder` but we are just trying to get a referral for `\\server\share` but interested to see what the real world problem was.

dHannasch:
> Thanks for the pickup here, are you able to share the scenario and error message you got when your encountered this error? From looking at the code it seems like it will fire if the DFS cache has an entry for `\\server\share\folder` but we are just trying to get a referral for `\\server\share` but interested to see what the real world problem was.

Ah, interesting. I had it happen when calling `smbclient.path.exists`.
https://gitlab.com/dHannasch/samba-test-cookiecutter/-/blob/master/%7B%7Bcookiecutter.samba_test_directory_name%7D%7D/test_samba.py#L48
```
Traceback (most recent call last):
  File "test_samba.py", line 107, in <module>
    test_access(default_context)
  File "test_samba.py", line 49, in test_access
    assert smbclient.path.exists(parent)
  File "/usr/lib/python3.8/site-packages/smbclient/path.py", line 27, in exists
    return _exists(path, False, True, **kwargs)
  File "/usr/lib/python3.8/site-packages/smbclient/path.py", line 141, in _exists
    stat(path, follow_symlinks=follow_symlinks, **kwargs)
  File "/usr/lib/python3.8/site-packages/smbclient/_os.py", line 550, in stat
    raw = SMBRawIO(path, mode='r', share_access='rwd', desired_access=FilePipePrinterAccessMask.FILE_READ_ATTRIBUTES,
  File "/usr/lib/python3.8/site-packages/smbclient/_io.py", line 291, in __init__
    tree, fd_path = get_smb_tree(path, **kwargs)
  File "/usr/lib/python3.8/site-packages/smbclient/_pool.py", line 264, in get_smb_tree
    referral = client_config.lookup_referral(path_split)
  File "/usr/lib/python3.8/site-packages/smbclient/_pool.py", line 150, in lookup_referral
    if idx > len(path_components) or referral_component != path_components[idx]:
IndexError: list index out of range
```
I couldn't figure out what chain of events was leading to the crash, but I figured "well, this line sure looks wrong".

The path I was trying to check on was of the form `\\server\share` and there does indeed exist an `\\server\share\folder`, so I think you have the right idea, though I don't really understand your comment.

(I might be missing a basic understanding of DFS. At the moment I'm just trying to brute-force the issue by figuring "well, observationally if I call `smbclient.listdir` on this magic path first, then everything works and I don't get STATUS_PATH_NOT_COVERED." I'm guessing from the word "referral" that something's going on under the hood selecting from among several SMB servers?)

jborean93:
Thanks for the info, the code you've changed definitely is the way to go as the original logic is wrong. This is just me merely trying to understand the scenario but that's not as important in this case.

> I might be missing a basic understanding of DFS

I don't blame you, DFS is just weird to me, just look at this fun chart from the Microsoft protocol docs https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dfsc/1ff7d611-ba19-49fb-95f4-7c5358b86834 :)

I may not follow this workflow exactly but what is currently there should work in most situations (excluding bugs like this). But essentially this is the workflow that is currently in place

* If `ClientConfig(domain_controller='mydc')` is set then a domain referral request is done
  * This step is totally optional and just helps when you can't resolve the domain name in a path to a DC
  * This domain referral request gets a list of domain names  in that DC, usually it's just the NETBIOS and dns name; `DOMAIN` and `domain.com`
  * These domain names are stored in the domain referral cache (important later)
* A request is made to access `\\domain.com\share\file.txt`
* We check the DFS referral cache to see if we've already accessed this DFS path
  * If we have a match on the cache we rewrite the path so we are talking to the DFS server
  * This is where it is failing for you due to the faulty cache lookup logic
* We check the domain referral cache to see if the first path component is a known domain
  * If we never set `domain_controller` in the client config this will never hit and we continue to the next step
  * If there was a domain referral match we send a DC referral request to get an order list of domain controllers in our domain
    * We store each domain controller in our domain referral cache for use later on
  * Now we know a prioritized list of DCs we send send a DFS referral request for `\\domain.com\share`
    * This gives us a list of DFS servers we can talk to to access that DFS share
    * This will rewrite our original path to `\\dfs-server\share\file.txt` for this request
    * The result is cached for later use
* We attempt to connect to `\\server\share` based on our path. This could be the origin `\\domain.com\share` or `\\dfs-server\share` depending on whether we had a cache match or sent the DFS referral request
  * If this fails with `STATUS_BAD_NETWORK_NAME` (share does not exist) then we send a DFS referral request to that host and cache the response
  * Now it will either fail again which is expected for a bad path or it will connect to the DFS server using the cached result

jborean93:
Thanks again for the fix.

dHannasch:
Thank you very much! I tried the `referral.target_hint.target_path`, and that's an actual domain. I think I understand a little more now, and I've updated https://gitlab.com/dHannasch/samba-test-cookiecutter/-/blob/master/%7B%7Bcookiecutter.samba_test_directory_name%7D%7D/test_samba.py to check on the DFS referral going forward.

...except that this domain is a completely different domain from the one that appears when I look at the shared drive in Windows (by right-clicking and opening Properties->DFS). So there must be another, completely different layer of indirection that I haven't found yet, where this domain forwards to the *real* domain that appears in Properties->DFS. (So I think in some sense what I'm calling `real_server_name` isn't the real server name, but I don't know it *what* sense.) But it's still worth knowing that *this* layer of indirection exists, since it's another thing to check on for debugging.

PR #93, opened 2021-04-26
Use a named function when checking whether a path is remote.

dHannasch:
Ultimately I want to extend `copytree` to allow copying to local, but as a first step towards that, this moves out the check for a remote path to a named function.

### DIFF
--- a/smbclient/shutil.py
+++ b/smbclient/shutil.py
@@ -62,7 +62,7 @@ from smbprotocol.structure import (
 )
 
 
-def is_remote_path(path):
+def is_remote_path(path):  # type: (str) -> bool
     """
     Returns True iff the given path is a remote SMB path (rather than a local path).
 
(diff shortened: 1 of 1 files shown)


smbclient/shutil.py
```diff
@@ -62,6 +62,16 @@
 )
 
 
+def is_remote_path(path) -> bool:
```
jborean93: For now you can do. I probably want to have 1 more release that supports 2.7

```suggestion
def is_remote_path(path):  # type: (str) -> bool
```

dHannasch:
Hmm. Codecov is reporting that this would somehow remove coverage for two error-condition lines in transport.py:
https://app.codecov.io/gh/jborean93/smbprotocol/compare/93/changes

I have no earthly idea how that's possible.

jborean93:
Must be some race condition like effect where sometimes certain code paths are excised there but others times it does now. The transport stuff is weird so it doesn't surprise me too much.

Issue #94, opened 2021-04-26
Dropping Python 2 support


dHannasch:
Should smbprotocol drop support for Python 2?
(https://www.python.org/doc/sunset-python-2/)

jborean93:
While I'm happy to drop Python 2.7 I probably won't drop it until there is a need to not actually use it. Right now I use it in a few places where Python 2.7 is being used so still being able to do bugfixes for that version is helpful.

jborean93:
Your wish has been granted with https://github.com/jborean93/smbprotocol/pull/104 :)

PR #95, opened 2021-04-26
Drop Python 2 support.

dHannasch:
This would remote support for Python 2 from smbprotocol (see https://github.com/jborean93/smbprotocol/issues/94).


jborean93:
This has been done with https://github.com/jborean93/smbprotocol/pull/104 which also dropped Python 3.5 as that is end of life.

PR #96, opened 2021-04-26
Join local or remote path

dHannasch:
This is built stop https://github.com/jborean93/smbprotocol/pull/98; see there first.

This creates a function that can join local or remote paths, and uses it in `copytree`, as a step towards allowing `dst` to be a local path.

After this, see https://github.com/jborean93/smbprotocol/pull/97.

### DIFF
--- a/smbclient/shutil.py
+++ b/smbclient/shutil.py
@@ -444,9 +444,15 @@ def _copy(src, dst, follow_symlinks, copy_meta_func, **kwargs):
     # Need to check if dst is a UNC path before checking if it's a dir in smbclient.path before checking to see if it's
     # a local directory. If either one is a dir, join the filename of src onto dst.
     if is_remote_path(ntpath.normpath(dst)) and isdir(dst, **kwargs):
+<<<<<<< HEAD
         dst = ntpath.join(dst, _basename(src))
     elif os.path.isdir(dst):
         dst = os.path.join(dst, _basename(src))
+=======
+        dst = ntpath.join(dst, basename(src))
+    elif os.path.isdir(dst):
+        dst = os.path.join(dst, basename(src))
+>>>>>>> a5a2ae4 (Single basename function for remote and local paths.)
 
     copyfile(src, dst, follow_symlinks=follow_symlinks)
     copy_meta_func(src, dst, follow_symlinks=follow_symlinks)
--- a/smbclient/shutil.py
+++ b/smbclient/shutil.py
@@ -331,7 +331,7 @@ def copytree(src, dst, symlinks=False, ignore=None, copy_function=copy2, ignore_
         if dir_entry.name in ignored:
             continue
 
-        src_path = ntpath.join(src, dir_entry.name)
+        src_path = join_local_or_remote_path(src, dir_entry.name)
         dst_path = join_local_or_remote_path(dst, dir_entry.name)
 
         try:
--- a/smbclient/shutil.py
+++ b/smbclient/shutil.py
@@ -38,7 +38,6 @@ from smbclient.path import (
     isdir,
     islink,
     samefile,
-    join_local_or_remote_path,
 )
 
 from smbprotocol import (
@@ -76,6 +75,21 @@ def _basename(path):
         return os.path.basename(path)
 
 
+def _join_local_or_remote_path(path, *paths):
+    """
+    Return the joined paths regardless of whether they are local or remote.
+    Currently, this only makes the decision based on the base path, and does not raise a ValueError if the additional
+    paths do not match.
+    :param path: The base path.
+    :param paths: The additional paths to append.
+    :return: Joined path.
+    """
+    if is_remote_path(path):
+        return ntpath.join(path, *paths)
+    else:
+        return os.path.join(path, *paths)
+
+
 def is_remote_path(path) -> bool:
     """
     Returns True iff the given path is a remote SMB path (rather than a local path).
@@ -83,7 +97,7 @@ def is_remote_path(path) -> bool:
     :param path: The filepath.
     :return: True iff the given path is a remote SMB path.
     """
-    return path.startswith('\\\\')
+    return path.startswith('\\\\') or path.startswith('//')
 
 
 def copy(src, dst, follow_symlinks=True, **kwargs):
@@ -331,8 +345,8 @@ def copytree(src, dst, symlinks=False, ignore=None, copy_function=copy2, ignore_
         if dir_entry.name in ignored:
             continue
 
-        src_path = join_local_or_remote_path(src, dir_entry.name)
-        dst_path = join_local_or_remote_path(dst, dir_entry.name)
+        src_path = _join_local_or_remote_path(src, dir_entry.name)
+        dst_path = _join_local_or_remote_path(dst, dir_entry.name)
 
         try:
             if dir_entry.is_symlink():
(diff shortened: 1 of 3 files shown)


dHannasch:
> Currently all the functions in smbclient/_os.py are purely for the SMB operations

(shrug) Having if you don't want the functions in `smbclient.path`, they're in `smbclient.shutil` now.

smbclient/shutil.py
```diff
@@ -63,6 +63,39 @@
...
+    """
+    Return the base name of pathname path.
+    This is the second element of the pair returned by passing path to the function split().
+    Note that the result of this function is different from the Unix basename program;
+    where basename for '/foo/bar/' returns 'bar', the basename() function returns an empty string ('').
+
+    :param path: The path to take the base of.
+    :return: The basename of the path.
+    """
+    if is_remote_path(path) or path.startswith(u"//localhost/share-encrypted/Pýtæs†-"):
```
jborean93, line 76: This sounds wrong, we shouldn't have code here to get the tests to pass. If the tests are failing without this 2nd check then the logic isn't sound and needs to be fixed. You can specify a path using either `/` or `\` as path separators and this library should be able to handle it. Essentially a remote path is one that starts with either `\\` or `//`.

smbclient/shutil.py
```diff
@@ -63,6 +63,39 @@
 )
 
 
+def basename(path):
```
jborean93, line 66: I don't think we should expose this officially yet, best to just prefix it with `_` if it's used internally here.

smbclient/shutil.py
```diff
@@ -63,6 +63,39 @@
...
+    :param path: The path to take the base of.
+    :return: The basename of the path.
+    """
+    if is_remote_path(path) or path.startswith(u"//localhost/share-encrypted/Pýtæs†-"):
+        return ntpath.basename(path)
+    else:
+        return os.path.basename(path)
+
+
+def join_local_or_remote_path(path, *paths):
```
jborean93, line 82: I don't think we should expose this officially yet, best to just prefix it with `_` if it's used internally here.

jborean93 (changes requested):

tests/test_smbclient_shutil.py
```diff
@@ -71,6 +74,14 @@ def _set_file_attributes(path, attributes):
             set_info(transaction, basic_info)
 
 
+def test_basename_local():
```
jborean93, line 77: It might be better to test one of the functions you are hoping would accept a local path instead of just testing this function directly. Same for join below.

dHannasch: I...guess? I sort of thought `copyfile()` already did accept local paths. The docstring says it does. I didn't think this particular PR actually changed any functionality. The point was just to extract the conditional from `_copy` to a reusable convenience function.

jborean93: This sort of ultimately leads into my comment here https://github.com/jborean93/smbprotocol/pull/106#issuecomment-834935960. There's a few PRs that are creating/moving helper functions with the ultimate goal of supporting remote <-> local copytree functionality. You are better off trying to do that one goal in 1 PR unless the amount of changes is quite drastic. It helps you to have 1 branch that tracks all your changes and to actually test these helper functions as part of what they are inteded for.

dHannasch:
> let's mark them as a draft PR when you open it until the required code already in flight has been merged in. That lets me know what reviews to focus on first.

O...kay? That was kind of the reason I put "This is built stop #98; see there first." as the first line in the description of this PR, and variations on that on others like https://github.com/jborean93/smbprotocol/pull/97 pointing here. To be honest I was a little confused why you were making comments on https://github.com/jborean93/smbprotocol/pull/97. I figured maybe you didn't like https://github.com/jborean93/smbprotocol/pull/98 since codecov was having that freakout. But it sounds like maybe it wasn't clear that #98 was the place to start?

The thing about marking this as a draft PR is that that doesn't indicate *which* other PR it's waiting on. I was thinking to give you bite-sized chunks, but lay out the entire sequence ahead of time so that, if you're happy with the first one and you merge it, you don't strictly have to wait on me to tell you what the next one is. I mean, for one thing, I might just not be around when you come by, decide you like a PR, and merge it. I wasn't around when you merged https://github.com/jborean93/smbprotocol/pull/93 or when you merged https://github.com/jborean93/smbprotocol/pull/99. If I use some non-specific way of noting that a PR depends on another PR, like mark-as-draft, then you could only ever merge one at a time, even if you would have been happy with the next one if you had only known which was the next one to look at. I guessed that your available time is probably bursty, and therefore when *I* have time I should try to lay out absolutely everything, so that (if there *aren't* changes needed) you can go through the whole stack without further input from me. I tend to assume, every time I post something here, that you might not actually see it for a week or more. (And on my end, I might be long gone when you reply, and who knows when I'd be back.)

> submit them as a single PR

As for *that*, I mean, obviously I can do that if you want. Obviously what I am in fact using on my end is all of these merged into a single branch. I just figured that a huge omnibus diff would be really difficult to read. Especially since a lot of the changes are logically orthogonal to each other.

Take https://github.com/jborean93/smbprotocol/pull/99.
On the one hand, I *do* have an ulterior motive: I want to `copytree` between local and remote. But if I gave you a PR and said "Here's a PR that makes copytree work on local directories" and you asked me "What the heck is all this where you're messing with `link()` and all these other functions, what does this have to do with copytree" I'd have no answer. Strictly speaking, https://github.com/jborean93/smbprotocol/pull/99 isn't necessary to make `copytree` work on local directories.

But obviously, as a question of plain causality, it *was* while working on `copytree` that I thought "Hey, you know what I really want to have here?" and came up with all the little functions like `is_remote_path`. Why? Sort of...trying not just to make `copytree` work now, but set it up so that it'll automagically keep working without any maintenance specific to `copytree`, just from the process of maintaining the other functions.

Like, I could have copy-pasted `.startswith('\\\\')` into `copytree()` and made it work that way. But suppose, to take a random example, that it later turned out `.startswith('\\\\')` was insufficient, that we also needed to test for paths starting with `//`. With the function extracted, that only needs to be changed in one place: https://github.com/jborean93/smbprotocol/pull/103
With the functionality duplicated, it would need to be changed in multiple places, but most importantly, the use I'd be *adding* in `copytree()` would be *one more place* that needed to be changed. A hypothetical version of `copytree()` that added local functionality *without* refactoring the rest of the codebase would become an ongoing maintenance problem. Does that make sense?

So I wanted to do some refactoring. But meanwhile, well, it's your code. Ultimately, you're the one that owns the darn thing, and any refactoring has to make changes to the codebase such that it comes out looking better *to you*. The idea is that extracting functions like `is_remote_path` is intended to look like a good idea not just for extending `copytree`, but for any similar functionality that we might want later. I was sort of hoping that these would be functions you'd want to use yourself, that you'd look at them and think "oh yeah, that's a good idea, next time I add a function and need to check whether a path is remote I'll call this `is_remote_path()` rather than inspect the path directly."

I'm trying to avoid having any more than the absolute minimum code *specifically for* the `copytree()` extension, if that makes sense. It's meant to be more "Hey, let's do some refactoring and whaddaya know, after that's done we can trivially extend `copytree` for free."

Each one of these just takes a piece of code that you already have, and extracts it into a named function for reuse. After taking all the bits of code that you already wrote and extracting them to make them accessible, the `copytree()` extension is just a trivial application of them.

jborean93:
> But it sounds like maybe it wasn't clear that #98 was the place to start?

You were clear, I was mostly trying to just have a glance over on all the PRs to get a general gist of where you were headed and seeing the same changes across them rather than just in the 1 PR was confusing me just a little bit. What annoys me about GitHub is once you merge one of those PRs the diff view is still showing that code as changed meaning it's more work for myself to see what those subsequent PRs are changing. This is a gripe I have on GitHub itself and not your work, in reality the changes made are pretty small so it's not the end of the world for me.

I definitely appreciative you are trying to make these changes in small piecemeal fashion but I think for me this scenario it might just be a bit too small :) Right now there are 7 PRs in place and I would need to go into each one to really see what I should be reviewing and when. This is mostly just me being lazy but I think in the case of adding this functionality into `copytree()` it might just be best to do it in one go.

dHannasch:
> seeing the same changes across them rather than just in the 1 PR was confusing me just a little bit. What annoys me about GitHub is once you merge one of those PRs the diff view is still showing that code as changed meaning it's more work for myself to see what those subsequent PRs are changing.

Ahhhh. Yes. Argh. That is something where you still need me to manually merge in master after you merge a previous PR, to refresh the displayed diff.

There are workarounds on GitLab, but...hmm. Yeah, I don't actually know a way to fix that on GitHub. Annoying.

(...darn it, there's gotta be a way to make that automatic, this is so basic, regular git on the command line displays the diff correctly after a subordinate branch has been merged into master...)

PR #97, opened 2021-04-26
Check whether remote path before calling makedirs

dHannasch:
This is built atop https://github.com/jborean93/smbprotocol/pull/96; look there first.

This allows `copytree` to work with a local destination.


jborean93:
For the same reasons listed here https://github.com/jborean93/smbprotocol/pull/100#issuecomment-834071301 this is not something I think `_os` should be handling. Any of the remote call `smbclient._os` or local call `stdlib.os` logic should be handled in `smbclient.shutil`, either in the function that requires it or in some other helper function.

jborean93:
My comment from https://github.com/jborean93/smbprotocol/pull/97#issuecomment-834073455 is still the ideal case. The `_os.py` functions are meant to be stubs for pure SMB calls. It's not meant to deal with cross SMB and local paths. The way forward for implementing local path support in `copytree` would be to handle the `is_remote_path` checks in `smbclient.shutil.py` that decide whether to call the `smbclient._os.py` function for SMB paths or the builtin `os/shutil` one.

There are quite a few merge conflicts that I encountered when trying to rebase https://github.com/jborean93/smbprotocol/pull/96 so I'm going to close this PR and if you still wish to implement this feature please open a new one with the changes you have in mind and we can move forward from there.

PR #100, opened 2021-04-27
Make scandir accept a local path

dHannasch:
This is built atop https://github.com/jborean93/smbprotocol/pull/99; look there first.

For https://github.com/jborean93/smbprotocol/blob/master/smbclient/shutil.py#L316.

Together with the others, this allows `copytree` to copy *from* a local directory as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+## 1.6.0 - TBD
+
+* Dropped support for Python 2.7 and 3.5
+* Add performance improvement on reads with large files
+
+
+## 1.5.1 - 2021-05-08
+
+* Unified DFS path handling when using any API that uses a transaction to open the file
+  * This includes `smbclient.rename` and `smbclient.replace`
+* Fixed up `smbclient.rename` to work with directories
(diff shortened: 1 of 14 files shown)


jborean93:
Sorry I started looking at your PRs and I got a bit sidetracked. Currently all the functions in `smbclient/_os.py` are purely for the SMB operations and I've tried to keep them separate from the `shutil` functions which is designed to give a filesystem agnostic/more human friendly layer over the top. If the end goal is to implement `copytree` with a local dir then these changes should occur in `copytree` itself.

This is similar to `copyfile` where [_os.copyfile](https://github.com/jborean93/smbprotocol/blob/af569b5f0a75e7eda36c4d12f0bbf921602edfc7/smbclient/_os.py#L134) is strictly to copy a file across the same SMB share (a specific operation in SMB). Compare that to [shutil.copyfile](https://github.com/jborean93/smbprotocol/blob/af569b5f0a75e7eda36c4d12f0bbf921602edfc7/smbclient/shutil.py#L104) which detects whether it should use the os copyfile or whether it needs to do a mixture of commands when the files aren't on the same SMB share.

jborean93:
See https://github.com/jborean93/smbprotocol/pull/107#issuecomment-887350202

Issue #91, opened 2021-04-15
Low read performance


mw-a:
Hi!

I'm looking into accessing files on SMB shares from Dask via fsspec and eventually smbprotocol. This works very well. But in tests it seemed that reading was a bit slow. Comparison with a Linux cifsfs mount and pysmbc showed that write performance is almost on par between them but reads with smbclient are only about half as fast as writes.

I've created this small testcase (and analogous ones for local/cifsfs and pysmb accesses):
```
import smbclient
import datetime

smbclient.register_session("dc")

string = ""
for j in range(1000):
    string += "%d:%s" % (j, ",".join("%d" % i for i in range(1000)))

start = datetime.datetime.now()
with smbclient.open_file("//dc/share/fsspec-smbprotocol.txt", "w", encrypt=False) as f:
    for i in range(100):
        f.write("%d:%s\n" % (i, string))

print("read: %s" % (datetime.datetime.now() - start))

start = datetime.datetime.now()
with smbclient.open_file("//dc/share/fsspec-smbprotocol.txt", "r", encrypt=False) as f:
    while f.read(1024*1024*10):
        pass

print("write: %s" % (datetime.datetime.now() - start))
```
In my testing environment it logs:
```
read: 0:00:02.762513
write: 0:00:04.584177
```
Is this something you're aware of or have an idea what could be causing it?

With Linux cifsfs mount and pysmbc both read and write of this ~380MiB test data is in the range of 2.5 to 3 seconds.

I've tried profiling the read and write parts and noticed that the read part spends a lot more time waiting for locks and, particularly, in https://github.com/jborean93/smbprotocol/blob/9a141ec4e1d2a60819c9c560a9629ef90e5a932d/smbprotocol/structure.py#L233.
Reproducer:
```
import cProfile, pstats, io
from pstats import SortKey
pr = cProfile.Profile()
pr.enable()

import smbclient

start = datetime.datetime.now()
with smbclient.open_file("//dc/share/fsspec-smbprotocol.txt", "r", encrypt=False) as f:
    while f.read(1024*1024*10):
        pass
print(datetime.datetime.now() - start)

pr.disable()
s = io.StringIO()
ps = pstats.Stats(pr, stream=s)
ps.sort_stats('cumtime')
ps.print_callers()
print(s.getvalue())
```
Output excerpt:
```
0:00:04.415829
   Ordered by: cumulative time

Function                                                                                                                            was called by...
                                                                                                                                        ncalls  tottime  cumtime
[...]

PR #102, opened 2021-05-07
Minor performance improvement on large reads

jborean93:
Thanks to @mw-a for the suggestion here. This adds a nice speed improvement when dealing with large SMB messages like a read response.

Fixes https://github.com/jborean93/smbprotocol/issues/91


jborean93:
Looks like this requires some more intrusive code for Python 2.7 which I'm not 100% comfortable doing right now. Will keep this open and merge once 2.7 is dropped after the next release.

PR #103, opened 2021-05-07
 or path.startswith('//')

dHannasch:


dHannasch:
codecov says that https://app.codecov.io/gh/jborean93/smbprotocol/compare/103/changes#D1L574 is somehow no longer covered although the lines after it, such as `query_flags = 0  # Only the first request should have set SMB2_RESTART_SCANS`, are still covered. I...don't know how that's even possible.

jborean93:
Looking at the output it seems like the lines have been shifted so the report is not showing the correct lines

image

There might be a bug on the codecov side trying to merge the diff (or lack of in this case) with the data it was given. I'll have to keep an eye on it. If you check the run for Azure Pipelines you can see that this file is unchanged and those lines are in fact covered https://dev.azure.com/jborean93/fa253176-5e40-44b7-b0e1-76408675a0b9/_apis/test/CodeCoverage/browse/13441289/Code%20Coverage%20Report_948/._py.3.html.

In reality the error here is the `smbprotocol/transport.py` has reduced coverage but the code in there is very variable to race conditions hence why it sometimes fires and other times it does not. I'll probably just need to add some `#pragma: nocover` on those affected lines to avoid this happening in the future.

jborean93 (approved):

PR #107, opened 2021-05-08
Allow copytree to work on local directories.

dHannasch:


jborean93:
There are a few merge commits and changes to these files recently. I'm going to close the PR until you find some time to rebase all the changes and get things in place for review. One thing I would note is to not expose `scandir` here for now. Just keep things internal like we've done for `copyfile` https://github.com/jborean93/smbprotocol/blob/master/smbclient/shutil.py#L152-L165.

PR #110, opened 2021-06-01
Ignore ENOTCONN socket errors

Alex-Weatherhead:
Address the (presumably) Linux-specific race condition in socket.shutdown()

See https://github.com/FreeOpcUa/python-opcua/issues/739 for a similar issue/fix


trevoryao:
Have also experienced this bug before, simple fix that would get rid of a lot of dirty work arounds

jborean93 (approved):
Thanks for the PR, dealing with sockets has been nothing but a pain so thank you for picking it up and submitting the fix.

PR #113, opened 2021-06-23
Few improvements

stastnypremysl:
Hello.

I have made few small improvements to your project.

Would it be possible to bump up a version, if you like my changes, so I can download it directly from PyPi?

Thanks
P.


jborean93:
Thanks for the PR. I had to make a slight tweak as `is_dir` isn't a function in `_os.py`. Instead you should be using `os.path.isdir` which returns whether the path is a bool or not.

PR #122, opened 2021-07-26
Fix comment to clarify requirement (dst argument, not src)

malthe:
This seems to make more sense.


jborean93 (approved):
Thanks for the pickup!

PR #130, opened 2021-07-28
Fix output (replace print())

MichaelKorn:
Just tried the new release and wondered about an unstructured output in my logs.


jborean93:
Crap, should have been more careful what I committed.

jborean93:
Sorry for the mess up here and thanks for fixing. I've just published [1.6.1 on PyPI](https://pypi.org/project/smbprotocol/1.6.1/) including this change.

MichaelKorn:
@jborean93, thanks for the immediate release!

Issue #117, opened 2021-07-15
listdir() sometimes blocks program


MichaelKorn:
The use of listdir() sometimes yields to a blocked program.

## Expected Behavior
1.  There should be a reasonable timeout for a listdir() call.
2. `TimeoutError: [Errno 110] Connection timed out` should be handled.

## Current Behavior
The use case:
- The application checks smb share folder (with listdir()) more than 3 times per second.
- The connection to the smb share is slow and maybe not completely stable (over VPN).
  - It is possible that a part of the communication between the server and client get lost.
- This application runs within a docker container (python:3.9-slim) with smbprotocol==1.5.1.

Sometimes the application hangs during the call of smbclient.listdir(). Until that point, the application usually runs for many hours, sometimes a few days.
During such failure state I retrieved debug information with faulthandler.dump_traceback(all_threads=True) (I removed the output for some Threads (DB connection and debugging)):
```
Thread 0x00007f00eeffd700 (most recent call first):
  File "/usr/local/lib/python3.9/site-packages/smbprotocol/transport.py", line 140 in _recv
  File "/usr/local/lib/python3.9/site-packages/smbprotocol/transport.py", line 119 in recv
  File "/usr/local/lib/python3.9/site-packages/smbprotocol/connection.py", line 1119 in _process_message_thread
  File "/usr/local/lib/python3.9/threading.py", line 892 in run
  File "/usr/local/lib/python3.9/threading.py", line 954 in _bootstrap_inner
  File "/usr/local/lib/python3.9/threading.py", line 912 in _bootstrap

Thread 0x00007f01008fb740 (most recent call first):
  File "/usr/local/lib/python3.9/threading.py", line 312 in wait
  File "/usr/local/lib/python3.9/threading.py", line 574 in wait
  File "/usr/local/lib/python3.9/site-packages/smbprotocol/connection.py", line 852 in receive
  File "/usr/local/lib/python3.9/site-packages/smbprotocol/open.py", line 1219 in _create_response
  File "/usr/local/lib/python3.9/site-packages/smbclient/_io.py", line 314 in commit
  File "/usr/local/lib/python3.9/site-packages/smbclient/_io.py", line 465 in open
  File "/usr/local/lib/python3.9/site-packages/smbclient/_io.py", line 410 in __enter__
  File "/usr/local/lib/python3.9/site-packages/smbclient/_os.py", line 242 in listdir
  File ..... my call of smbclient.listdir()
```
[...]

PR #135, opened 2021-09-08
Connection health checks

MWedl:
This PR implements health checks for server connections to prevent infinite blocks. 

When the server connection gets lost and no data is received anymore, the receiving thread waits infinitely. After 10 minutes a `TimeoutError` is raised in `transport.recv()` and a keepalive message gets sent to the server.
However, there were no checks in place if the server actually answers the keepalive echo requests, which lead to an infinite timeout loop.

My implementation checks on timeouts, whether unanswered keepalive echo requests exist. These keepalives are only sent during (previous) timeouts. When such keepalives exist, that means the server did not respond within `2 * timeout`. Then we assume that the connection is dead and close it.

 Relates https://github.com/jborean93/smbprotocol/issues/117

### DIFF
--- a/smbprotocol/connection.py
+++ b/smbprotocol/connection.py
@@ -83,7 +83,6 @@ from smbprotocol.structure import (
 )
 
 from smbprotocol.transport import (
-    _TimeoutError,
     Tcp,
 )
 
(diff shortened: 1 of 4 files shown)


jborean93 (approved):

jborean93:
Thanks for the PR. While I understand the issue is still present this should hopefully stop any long running hangs and alleviate some of the immediate concerns.

PR #138, opened 2021-09-18
Added missing symbols and functionality

kvv81:
We use smbprotocol for SMB2 server testing and we need some basic functionality missed in current code


jborean93 (commented):
Looks good, is it possible to add an exception class for each of the NtStatus values you've added? See https://github.com/jborean93/smbprotocol/blob/master/smbprotocol/exceptions.py#L284 and below for how I've done it before. The messages come from https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55 and the exception class name is just the PascalCase form of the NtStatus value without `STATUS_`.

kvv81:
Done, please re-review 1st commit.

jborean93 (approved):

jborean93:
Thanks for the PR, I've pushed a change that adds the changelog entry but everything else looks great.

PR #140, opened 2021-09-20
Support not following symlinks in SMBDirEntry.from_path

MWedl:
`stat()` supports `follow_symlinks=False` passed in `kwargs`, but `SMBRawIO(path, **kwargs)` does not and raises an Exception.

### DIFF
--- a/smbclient/_os.py
+++ b/smbclient/_os.py
@@ -1219,8 +1219,8 @@ class SMBDirEntry(object):
             return self._lstat
 
     @classmethod
-    def from_path(cls, path, **kwargs):
-        file_stat = stat(path, **kwargs)
+    def from_path(cls, path, follow_symlinks=True, **kwargs):
+        file_stat = stat(path, follow_symlinks=follow_symlinks, **kwargs)
 
         # A DirEntry only needs these 2 properties to be set
         dir_info = FileIdFullDirectoryInformation()


jborean93:
I'm not opposed to the change I'm just curious what the use case is of calling `from_path` here? It was only really added originally for `smbclient.shutil.copytree` to support a use case it had.

MWedl:
I'm building a SMB share crawler which performs some permission and content checks. One of the main features is that crawling can be stopped and continued where it left off. Encountered files are also periodically re-checked to see if something changed. Therefore all paths are stored in a small SQLite DB. When visiting the path, `SMBDirEntry.from_path()` is called. Files, directories and symlinks are handled differently by the crawler.
While testing I encountered a symlink which points to a local file. Because of that, a `SMBLinkRedirectionError` is raised in `SMBDirEntry.from_path()` as symlinks are always followed. This was before I could even check if the path is a symlink or not.

jborean93:
That's fair, if you are building the stat info from a path is there a reason why you aren't just calling `stat` or `lstat` manually? The main reason why `SMBDirEntry` exists is for `scandir` to provide a more efficient dir enumeration result with the ability to get more information as required by then calling `stat`. If you already have the path you can bypass this by calling `stat` directly.

The changes here are fine just trying to figure out the use case here and whether there's something missing in another function.

jborean93 (approved):

MWedl:
The reason why I am using `SMBDirEntry` instead of raw `stat()` results, are the methods `is_dir()`, `is_file()` and `is_symlink()`.

PR #141, opened 2021-09-22
Added high-level 'Client' concept, tests code refactoring

kvv81:
Effort for code simplicity, redability and no duplications (when suitable).
Code improvement for complex use-cases when several parallel clients are used.

-Added smbclient.Client class
-Extracted per-instance logic of global ClientConfig into ClientConfigInstance
-Added new RealClientConfig class as a descendant of ClientConfigInstance
-Fixed 'smb_real' fixture to return RealClientConfig instead of tuple
-Refactored the code of tests to use smbclient.Client when suitable and
    use 'smb_real.attribute' instead of 'smb_real[number]'

Signed-off-by: Volodymyr Khomenko <khomenko.volodymyr@gmail.com>


kvv81:
The main idea of this change is to make the code compact and simple for happy path.
In the most cases you need only one connection, session and tree so lower-level objects of Connection, Session and TreeConnect take a lot of efforts for init/destroy. We need higher-level concept of 'Client' to use this abstraction for basic flows. For sure, in some cases (when you need several sessions per the same connection or several trees for the same session) you still need to use lover-level objects - as before.

Old abstraction level of smbclient._os  POSIX-like functions (like mkdir and stat) are NOT suitable in some cases - when we want to use native Windows-like client flow. Also the mechanism of cached objects in pool makes this way less reusable - for example when we need two separate connections to the same server (but caching logic forces us to reuse the same connection).

I can propose the particular example we have - testing SMB parallel leases break flow (in our SMB server this code is very tricky).
In this use-case, three different clients have separate connections to the same server/share and they want to operate with the same file.
The first client opened the file-handle for this file with RHW lease (ReadHandleWrite, i.e. all possible client-side caching with exclusive access to file). After that two other concurrent clients want to open the same file with maximal possible lease level and some access_mask that may conflict with caching (for example, if some client want to Write the file, Read caching by other clients is not possible anymore and lease bust be broken to 'None' level).

Internally, on the reception of each conflicting Create request the server must send Lease break notification to the 1st client who still holds a cached state for it. And also the race between 2nd and 3rd client may occur so lease break notification can be sent several times to reduce the level of lease: RHW -> RH -> R -> None.

Without using higher-level concept of Client this code will be too messy (you need 12 different objects here - 3 Connections, 3 Sessions, 3 TreeConnects and 3 Opens).

Issue #136, opened 2021-09-09
Memory leaks with DFS


mrmaxi:
I try to use smbclient for read DFS-share but realise that memory leaks occurs.
I use docker image python:3.8-slim with installed smbprotocol 1.6.2 (i check on 1.5.1 too, bug still exists).
I connect to dfs share and listdir in loop, but memory usage grow little by little. 
In simple test script:
```python
from time import sleep
import smbclient

from local_settings import in_dir, smbLogin, smbPassword, smbDomainController
smbclient.ClientConfig(username=smbLogin, password=smbPassword, domain_controller=smbDomainController)

while True:
    files = smbclient.listdir(in_dir)
    print(files)
    sleep(0.5)
```

 **it grow from 20 Mb upto 200Mb in half an hour**.

This is tracemalloc output after 30 minutes reading DFS-share:
```python
/usr/local/lib/python3.8/site-packages/smbprotocol/structure.py:73: size=13.0 MiB, count=46175, average=296 B
/usr/local/lib/python3.8/site-packages/smbprotocol/dfs.py:378: size=6546 KiB, count=99414, average=67 B
/usr/local/lib/python3.8/site-packages/smbprotocol/structure.py:356: size=2893 KiB, count=46232, average=64 B
/usr/local/lib/python3.8/site-packages/smbprotocol/structure.py:537: size=2580 KiB, count=14202, average=186 B
/usr/local/lib/python3.8/site-packages/smbprotocol/structure.py:74: size=2536 KiB, count=8774, average=296 B
/usr/local/lib/python3.8/site-packages/smbprotocol/structure.py:810: size=2210 KiB, count=31424, average=72 B
/usr/local/lib/python3.8/encodings/utf_16_le.py:16: size=1355 KiB, count=21307, average=65 B
/usr/local/lib/python3.8/site-packages/smbprotocol/dfs.py:390: size=1276 KiB, count=14202, average=92 B
/usr/local/lib/python3.8/site-packages/smbprotocol/structure.py:216: size=669 KiB, count=8643, average=79 B
/usr/local/lib/python3.8/site-packages/smbprotocol/dfs.py:267: size=633 KiB, count=12789, average=51 B
```

And this is `docker stats` output for DFS-share:
```bash
CONTAINER ID   NAME      CPU %     MEM USAGE / LIMIT     MEM %     NET I/O          BLOCK I/O   PIDS
[...]

PR #144, opened 2021-10-17
fix: remove expired referrals from dfs cache

mrmaxi:
Fix memory leak with DFS https://github.com/jborean93/smbprotocol/issues/136

### DIFF
--- a/smbclient/_pool.py
+++ b/smbclient/_pool.py
@@ -148,10 +148,14 @@ class ClientConfig(object, metaclass=_ConfigSingleton):
             if domain.domain_name.lower() == ("\\" + domain_name.lower()):
                 return domain
 
+    def clear_expired_cache(self):
+        self._referral_cache = [refferal for refferal in self._referral_cache if not refferal.is_expired]
+
     def lookup_referral(self, path_components):  # type: (List[str]) -> Optional[ReferralEntry]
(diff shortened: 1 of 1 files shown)


smbclient/_pool.py
```diff
@@ -148,10 +148,14 @@ def lookup_domain(self, domain_name):  # type: (str) -> Optional[DomainEntry]
             if domain.domain_name.lower() == ("\\" + domain_name.lower()):
                 return domain
 
+    def clear_expired_cache(self):
```
jborean93, line 151: Thanks for working on this, sorry I haven't gotten to it myself. 1 minor nit:

* Can we make this `_clear_expired_cache` as it shouldn't really be something that is used outside of the internal lookups

jborean93: Yea I merged it in because I assumed I was going to push my own PR that renamed it and also did the expiry check for the domain controller lookup but unfortunately I was held up with some other stuff. Feel free to submit another PR, otherwise I'll get to it sometime this week.

> But the other side, maybe it's premature optimization, what do you think?

I think what you have here is fine, it will weed out the expired referrals when a lookup occurs and I would hope getting the expiry time wouldn't be too hard. I would say unless it's a known problem we probably don't need any further optimisations here.

PR #153, opened 2021-11-04
InfoAdditionalInformation should be of type flags and not enum fixes #152

### LINKED ISSUE #152: additional_information in SMB2SetInfoRequest is set as enum despite Microsoft documentation stating it should support flags
The `additional_information` field in `SMB2SetInfoRequest` (https://github.com/jborean93/smbprotocol/blob/master/smbprotocol/open.py#L1040) is defined as an `EnumField` despite Microsoft documentation ([MS-SMB2], section 2.2.39 SMB2 SET_INFO Request) stating that `additional_information` is a 4-byte bit field of **flags**. 

For example, setting `additional_information` to `OWNER_SECURTIY_INFORMATION | GROUP_SECURITY_INFORMATION | DACL_SECURITY_INFORMATION` is currently impossible throwing an `Enum value 7 does not exist in enum type` exception.

wokis:
…#152

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 1.8.3 - TBD
+
+* Set `additional_information` field in `SMB2SetInfoRequest` to a flags field rather than a single enum as per protocol docs
+
+
 ## 1.8.2 - 2021-11-03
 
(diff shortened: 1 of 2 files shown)


jborean93:
Thanks for the PR, just added a changelog entry but the changes look good.

PR #162, opened 2022-01-17
Fix incorrect name mangling

linuskr:
Since `self.__kwargs` is set in `SMBRawIO.__init__`, its name will always be `_SMBRawIO__kwargs`.
This fixes connection_cache not being passed to get_smb_tree in dfs referrals


jborean93 (approved):

jborean93:
Thanks for the pickup on this.

Issue #85, opened 2021-03-25
replace/rename support for DFS file path


mrmaxi:
Thanks a lot for so great work!
I try to use your library with DFS file shares, but get some trouble.
"open_file", "listdir", "walk" and "remove" work correctly, but "rename" and "replace" cause STATUS_ACCESS_DENIED exception.
```
>>>smbclient.ClientConfig(username='user', password=’xxxx')

Traceback (most recent call last):
  File "C:/APPS/auto_env38/smbprotocol/t1.py", line 8, in <module>
    smbclient.replace('\\\\my-company.com\\DFS\\Report$\\Used\\Test\\Used\\FRB\\4.txt', '\\\\ my-company.com\\DFS\\Report$\\Used\\Test\\Used\\FRB\\6.txt')
  File "C:\APPS\auto_env38\smbprotocol\smbclient\_os.py", line 488, in replace
    _rename_information(src, dst, replace_if_exists=True, **kwargs)
  File "C:\APPS\auto_env38\smbprotocol\smbclient\_os.py", line 1047, in _rename_information
    set_info(transaction, file_rename)
  File "C:\APPS\auto_env38\smbprotocol\smbclient\_io.py", line 237, in __exit__
    self.commit()
  File "C:\APPS\auto_env38\smbprotocol\smbclient\_io.py", line 277, in commit
    raise failures[0]
smbprotocol.exceptions.SMBOSError: [Error 0] [NtStatus 0xc0000022] Unknown NtStatus error returned 'STATUS_ACCESS_DENIED': '\\my-company.com\DFS\Report$\Used\Test\Used\FRB\4.txt'
```

If I open_file().read() first, and then replace, it cause another exception “smbprotocol.exceptions.SMBOSError: [Error 2] [NtStatus 0xc000003a] No such file or directory”
```
print(smbclient.open_file('\\\\my-company.com\\DFS\\Report$\\Used\\Test\\Used\\FRB\\4.txt').read())
smbclient.replace('\\\\my-company.com\\DFS\\Report$\\Used\\Test\\Used\\FRB\\4.txt', '\\\\my-company.com\\DFS\\Report$\\Used\\Test\\Used\\FRB\\6.txt')

Hello world
Traceback (most recent call last):
  File "C:/APPS/auto_env38/smbprotocol/t1.py", line 8, in <module>
[...]

PR #171, opened 2022-03-20
Dfs refresh referrals

### LINKED ISSUE #170: STATUS_ACCESS_DENIED when write by dfs path before read

mrmaxi:
fix #170 - STATUS_ACCESS_DENIED when write by dfs path before read

Remember that path if DFS if it discovered once, and than when referral was expired only refresh it.
Use persistent _referral_cache as dict, so when referral is expired just refresh it (instead of removing).

### DIFF
--- a/smbclient/_pool.py
+++ b/smbclient/_pool.py
@@ -278,11 +278,15 @@ def get_smb_tree(path, username=None, password=None, port=445, encrypt=None, con
 
     # Check if we've already got a referral match for the path specified and use that path instead.
     referral = client_config.lookup_referral(path_split)
+    expired_referral = None
     if referral and not referral.is_expired:
         path = path.replace(referral.dfs_path, referral.target_hint.target_path, 1)
         path_split = [p for p in path.split("\\") if p]
 
     else:
+        # Save expired referral for refresh it
+        expired_referral = referral
+
         # If there was no referral match, check if the hostname portion matches any known domain names
         domain_referral = client_config.lookup_domain(path_split[0])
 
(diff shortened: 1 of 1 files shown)


jborean93:
Nice picking up on your original problem, I'll have to play around with it to see if I can replicate it as it's certainly not something I would expect. I think we might need a different approach in the PR as keeping the referrals around resulted in a memory bloat reporting in https://github.com/jborean93/smbprotocol/issues/136. Maybe we should keep the existing cache and the expiry checks but have another simple list that contains paths known to be DFS paths as part of the check. That way it still knows to do the DFS referral request even on an expiry but the referral responses aren't stored indefinitely.

mrmaxi:
> Maybe we should keep the existing cache and the expiry checks but have another simple list that contains paths known to be DFS paths as part of the check. That way it still knows to do the DFS referral request even on an expiry but the referral responses aren't stored indefinitely.

I thought about this, but in this case we need to duplicate lookup_referral code for find if requested path if dfs. And I don't found simple solution.

jborean93:
I've taken some time to look at this further and while this might alleviate the problem temporarily for you the issue still remains where the first request is going to error with `STATUS_ACCESS_DENIED` causing smbclient to raise the exception. Without the server actually responding with the expected errors to start the DFS process anything on top is going to be a hack really. At a guess your domain controller actually has a share or path that the code is legitimately trying to open, or failing that maybe the DFS share permissions themselves cause it to only work if opening with read access or when being requested to the proper DFS server.

It is due to this problem I see the proper solution is ensuing you've resolved the proper DFS servers for your domain rather than rely on the rudimentary check that `smbclient` tries to do for you without knowing what your domain controller is called.

As mentioned in https://github.com/jborean93/smbprotocol/issues/85#issuecomment-810841141 and further down that thread, in these cases you should be setting the following at the start of your script to ensure the domain referrals are looked up properly.

```python
import smbclient

smbclient.ClientConfig(domain_controller='dc.hostname.realm')
```

You can add `username` and/or `password` if you can't rely on implicit Kerberos authentication but otherwise what this code does is:

* Connects to the `domain_controller` specified
* Does a DFS referral request to get a list of domain names known by that DC
* Caches these domain names for future requests

When an actual request is seen to connect to one of these now known domain names it knows that this is for a DFS path and can do the lookup pre-emptively avoiding this problem altogether. The reason why you don't see this problem when running it natively on Windows is that it already knows the domain controller to contact and the domain information. On Linux, and in this pure Python implementation it does not have that luxury so it relies on what the caller has provided or error codes returned back. In this case it gets back a `STATUS_ACCESS_DENIED` so assumed the path is legitimate but you don't have access to it. Why it returns `STATUS_ACCESS_DENIED` rather than one of the expected error codes to start the lookup I don't have the slightest idea why and I cannot seem to replicate it myself. If this was working then this wouldn't be an issue because the DFS fallback code would run and check the cache or resend the DFS request as needed.

I apologise for getting back to you so late on the PR but unfortunately this is a workaround that keeps around expired caches causing a leak over time for something that can be solved by setting `smbclient.ClientConfig(domain_controller='...')` and letting smbclient know for sure what is a DFS request or not initially.

PR #184, opened 2022-06-30
Allow server side copy across different shares on the same server

coreyleavitt:


jborean93:
Sorry for such a delay in looking at this, I've tested the changes and they look good to me. Thanks for fixing this.

PR #212, opened 2023-03-23
Python3 style updates

tik-stbuehler:
Given "Python 3.7+" is listed as requirement, there are a few things that could be updated.
Also cleanup a previous "black" run that removed linebreaks between string literals but didn't merge them.

Regarding the `super(...)` cleanup: there was one non-trivial use in `_SMBErrorRegistry.__call__` which I replaced with `type.__call__`. Perhaps this method should start with `if cls != SMBResponseException: return super().__call__(header)` anyway - or ensure the status code of a passed header matches the `cls._STATUS_CODE` for subclasses (or even forbid passing a header to subclasses).

(The git history looks a little bit like you squash PRs to merge them? I'd appreciate it if you don't do that with mine - I split my changes into commits on purpose, and I'm happy to rebase/edit them :) )

### DIFF
--- a/build_helpers/check-smb.py
+++ b/build_helpers/check-smb.py
@@ -28,7 +28,7 @@ if __name__ == "__main__":
             print("Connection attempt %d failed: %s" % (attempt, str(e)))
             attempt += 1
             if attempt == total_attempts:
-                raise Exception("Timeout while waiting for SMB server to come " "online")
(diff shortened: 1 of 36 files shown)


tik-stbuehler:
I didn't run black/isort on `.` (because it went crazy on my `venv` directory), only on `src` - now also added `tests`.

Issue #216, opened 2023-06-09
Passing initial token to spnego for ntlm auth


maciejwie:
Hello,
I'm talking to an older IBM i server which can be picky about what it connects to. In particular, it rejects connections where the workstation name is null. This works on Windows because the underlying SSPI library always sets the `Version` flag and and sends the workstation name. On Linux, the `spnego` library checks the `in_token` for the `Version` flag and omits those sections if it's not found. In my testing, if I force enable sending that section then everything works.

What do you think about adding an ability to pass some parameters through to `spnego`? The "nicest" way would be to take in just the flags, and the more generic way would be to take the encoded bytes of the token itself.

maciejwie:
For what it's worth, I patched my local `pyspnego` to always set the `workstation` field as a workaround so I'm not sure if this PR is the way to do it either. My thinking was was `pyspnego` has a mechanism for forcing the field to be present, so it shouldn't be the one changing, and only noticed later that you have authored both which makes choosing the "more correct" place a lot easier. I defer to your thoughts here.

Capturing the logs at DEBUG level, it looks like the 2nd SMB2_SESSION_SETUP message to me. I've redacted the step input an output to avoid leaking things when I don't understand the format, but can pull things out if it will help. In this capture, the `auth_protocol` is the default `negotiate` and running in a Debian Docker container.
```
INFO:smbprotocol.connection:Negotiated dialect: (514) SMB_2_0_2
INFO:smbprotocol.connection:Connection require signing: True
INFO:smbprotocol.session:Initialising session with username: MY_DOMAIN\MY_USERNAME
DEBUG:smbprotocol.session:Decoding SPNEGO token containing supported auth mechanisms
DEBUG:spnego._negotiate:SPNEGO step input: [step input 1 redacted]
DEBUG:spnego._negotiate:Attempting to create ntlm context when building SPNEGO mech list
DEBUG:spnego._negotiate:SPNEGO step output: [step output 1 redacted]
INFO:smbprotocol.session:Sending SMB2_SESSION_SETUP request message
INFO:smbprotocol.session:Receiving SMB2_SESSION_SETUP response message
DEBUG:smbprotocol.transport:Socket recv() returned 4 bytes (total 4)
DEBUG:smbprotocol.transport:Socket recv(357) (total 357)
DEBUG:smbprotocol.transport:Socket recv() returned 357 bytes (total 357)
DEBUG:smbprotocol.transport:Socket recv(4) (total 4)
[...]

PR #217, opened 2023-06-09
Allow initial token to be set for NTLM

maciejwie:
When using NTLM authentication, the in_token passed to the spnego context.step() method is always None. However, this is the only mechanism which allows setting SMB negotiation flags.

Add optional initial_token parameter to be used when auth_method is 'ntlm' which allows this configuration to flow to spnego.

Addresses #216


maciejwie:
This is superseded by https://github.com/jborean93/pyspnego/pull/65

Issue #221, opened 2023-06-23
Additional features in shutil.py


brno32:
My team is finding we're writing a lot of wrapper code that could probably just be handled in shutil.py

For example, if `copytree` supported remote to local and local to remote operations instead of just remote to remote, this would take care of most of our needs.

Could we add support?

jborean93:
Sure, `shutil` is meant to be helpers on top of the core SMB protocols which would include copies on remote <-> local. Just needs someone to add a PR with tests.

PR #227, opened 2023-06-27
adding support for local to remote operations in shutil.py::copytree

brno32:
pt. 2 of addressing https://github.com/jborean93/smbprotocol/issues/221

Tackling symlinks looks a little daunting, so I'd like to circle back after making sure my current approach is aligned with the project.

It crossed my mind to have both `LocalDirEntry` and `SMBDirEntry` inherit from the same abstract class, but I've opted to keep it simple

### DIFF
--- a/src/smbclient/_os.py
+++ b/src/smbclient/_os.py
@@ -1172,6 +1172,24 @@ def _set_basic_information(
         set_info(transaction, basic_info)
 
 
+class LocalDirEntry:
+    """Mimics the structure of SMBDirEntry, but instead encapsulates a directory/file on a local filesystem"""
+
+    def __init__(self, path) -> None:
+        self.path = path
+
+    @property
+    def name(self):
+        """The entry's base filename, relative to the os.listdir() path argument."""
+        return os.path.basename(self.path)
+
+    def is_symlink(self):
+        return os.path.islink(self.path)
+
+    def is_dir(self):
+        return os.path.isdir(self.path)
+
+
 class SMBDirEntry:
     def __init__(self, raw, dir_info, connection_cache=None):
         self._smb_raw = raw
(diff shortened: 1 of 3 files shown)


jborean93:
Hey sorry I haven't had a chance to look at this yet, I've got a few things on my plate right now.

src/smbclient/_os.py
```diff
@@ -1172,6 +1172,24 @@ def _set_basic_information(
         set_info(transaction, basic_info)
 
 
+class LocalDirEntry:
```
jborean93, line 1175: I would probably put this in `shutil.py` as it's not SMB related but rather a shim used by the shutil function. I'm trying to keep `_os.py` strictly related to SMB operations. I'm also wondering if it's even needed as we could just utilise the result of [os.scandir](https://docs.python.org/3/library/os.html#os.scandir) which returns a [DirEntry](https://docs.python.org/3/library/os.html#os.DirEntry) object I modeled `SMBDirEntry` from.

src/smbclient/shutil.py
```diff
@@ -295,7 +293,10 @@ def copytree(
     :param kwargs: Common arguments used to build the SMB Session for any UNC paths.
     :return: The dst path.
     """
-    dir_entries = list(scandir(src, **kwargs))
+    if is_remote_path(src):
+        dir_entries = list(scandir(src, **kwargs))
+    else:
+        dir_entries = [LocalDirEntry(os.path.join(src, result)) for result in os.listdir(src)]
```
jborean93, line 299: using `os.scandir` is most likely going to be more efficient and should theoretically remove the need for `LocalDirEntry`.

jborean93 (approved):

jborean93:
Sorry it took so long to review this.

PR #233, opened 2023-08-12
Do not setup logging in libraries

dolfinus:
I found that your library modifies logger settings.

This should be done explicitly by user or application which uses the library, but never library itself.


jborean93 (approved):
Sorry for taking so long to get back to this one. The changes look good.

PR #234, opened 2023-08-12
Upgrade to Python 3.7+ syntax

dolfinus:
* Run `pyupgrade --py37-plus ./**/*.py` to upgrade Python 2.7 syntax to 3.7+
* Replace `%s` and `.format()` syntax with f-strings
* For logging use `%s` string syntax, but without explicit `"msg" % (variables,)` - let `logging` do this if logging is enabled for a specific level
* Convert `# type` comments to type hints, replace `List[type]`, `Optional[type]` and `Union[type1, type2]` with `list[type]`, `type | None` and `type1 | type2` (using `from __future__ import annotations`).
* Use `os.fspath` instead of `str` to convert Path-like objects to string path representation
* Update examples to use logging instead of `print`


src/smbprotocol/exceptions.py
```diff
@@ -201,11 +197,11 @@ class SMBResponseException(SMBException, metaclass=_SMBErrorRegistry):
 
     _BASE_MESSAGE = "Unknown error."
 
-    def __init__(self, header):  # type: (SMB2HeaderResponse) -> None
+    def __init__(self, header: SMB2HeaderResponse):
         self.header = header
 
     @property
-    def error_details(self):  # type: () -> List[any]
+    def error_details(self) -> list[Any]:
```
jborean93, line 204: Still needs typing.List

src/smbprotocol/session.py
```diff
@@ -272,7 +271,7 @@ def connect(self):
                 protocol=self.auth_protocol,
             )
         except spnego.exceptions.SpnegoError as err:
-            raise SMBAuthenticationError("Failed to authenticate with server: %s" % str(err.message))
+            raise SMBAuthenticationError("Failed to authenticate with server: %s", err.message)
```
jborean93, line 274: I think you still want `%` here or to convert it to an f string. Unless exceptions support such a thing?

src/smbprotocol/session.py
```diff
@@ -282,7 +281,7 @@ def connect(self):
             try:
                 out_token = context.step(in_token)
             except spnego.exceptions.SpnegoError as err:
-                raise SMBAuthenticationError("Failed to authenticate with server: %s" % str(err.message))
+                raise SMBAuthenticationError("Failed to authenticate with server: %s", err.message)
```
jborean93, line 284: Ditto on the this exception.

src/smbprotocol/structure.py
```diff
@@ -161,6 +160,10 @@ def __str__(self):
     def __len__(self):
         return self._get_packed_size()
 
+    @property
+    def endian_prefix(self):
```
jborean93, line 164: Prefer to keep this as `_endian_prefix` just to make sure it's more of an internal property.

jborean93 (commented):
Thanks for working on this. While I believe the type annotations might be fine if the future annotations are used I would prefer to ensure the mypy checks for each supported Python version still work so we might have to still use some of the older `typing` constructs.

src/smbclient/_os.py
```diff
@@ -182,9 +181,8 @@ def copyfile(src, dst, **kwargs):
                     copychunk_response = SMB2SrvCopyChunkResponse()
                     copychunk_response.unpack(result)
                     if copychunk_response["chunks_written"].get_value() != len(batch):
-                        raise IOError(
-                            "Failed to copy all the chunks in a server side copyfile: '%s' -> '%s'"
-                            % (norm_src, norm_dst)
+                        raise OSError(
```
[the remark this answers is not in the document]
dolfinus: https://docs.python.org/3/library/exceptions.html#IOError
`The following exceptions are kept for compatibility with previous versions; starting from Python 3.3, they are aliases of [OSError](https://docs.python.org/3/library/exceptions.html#OSError).`

README.md
```diff
@@ -57,9 +57,9 @@ guarantee. To verify that Kerberos is available on Linux you can run the followi
 ```python
 try:
     from gssapi.raw import inquire_sec_context_by_oid
-    print("python-gssapi extension is available")
+    log.debug("python-gssapi extension is available")
```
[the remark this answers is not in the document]
dolfinus: Okay

src/smbclient/_pool.py
```diff
@@ -448,7 +448,7 @@ def reset_connection_cache(fail_on_error=True, connection_cache=None):
             if fail_on_error:
                 raise
             else:
-                warnings.warn("Failed to close connection %s: %s" % (name, str(e)))
+                log.exception("Failed to close connection %s", name)
```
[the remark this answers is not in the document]
dolfinus: Yes, this is that `log.exception` is designed for. But I've changed it to `log.warning(msg, exc_info=e)` to match previous behavior

src/smbprotocol/exceptions.py
```diff
@@ -47,7 +49,7 @@ class SMBOSError(OSError, SMBException):
 
     def __init__(self, ntstatus, filename, filename2=None):
         self.ntstatus = ntstatus
-        self.filename2 = str(filename2) if filename2 else None
+        self.filename2 = os.fspath(filename2) if filename2 else None
```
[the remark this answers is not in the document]
dolfinus: https://docs.python.org/3/library/os.html#os.fspath

It is recommended to convert file paths matching `os.PathLike` protocol (like `pathlib.Path`) to strings instead of using just `str(...)`

src/smbprotocol/session.py
dolfinus: Replaced with `raise SMBAuthenticationError(msg) from err`

src/smbprotocol/structure.py
dolfinus: Okay

src/smbclient/_pool.py
```diff
@@ -99,9 +98,9 @@ def __init__(
         self.skip_dfs = skip_dfs
         self.auth_protocol = auth_protocol
         self.require_secure_negotiate = require_secure_negotiate
-        self._domain_controller = domain_controller  # type: Optional[str]
-        self._domain_cache = []  # type: List[DomainEntry]
-        self._referral_cache = []  # type: List[ReferralEntry]
+        self._domain_controller: str | None = domain_controller
```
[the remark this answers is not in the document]
dolfinus: I've added `from __future__ import annotations` to the beginning of the file. This allows to use modern annotations syntax in old Python versions, like 3.7, by converting all type hints to string representation. So this new syntax does not affect runtime, as you can see, tests are not failing on Python 3.7.

Tools like mypy can parse type hints to get proper types.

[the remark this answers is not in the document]
jborean93: Thanks for the clarification, I always assumed mypy was still dependent on what Python actually supported. Good to know I can use the newer syntax because it is much nicer.

src/smbprotocol/session.py
```diff
@@ -272,7 +271,7 @@ def connect(self):
                 protocol=self.auth_protocol,
             )
         except spnego.exceptions.SpnegoError as err:
-            raise SMBAuthenticationError("Failed to authenticate with server: %s" % str(err.message))
+            raise SMBAuthenticationError("Failed to authenticate with server") from err
```
jborean93, line 274: While it's good to do the inner exception I would probably like to keep the original exception error inside the main exception string just for preserving how ti works today or if someone is catching that in their code and doing `str(e)` themselves.

src/smbprotocol/session.py
```diff
@@ -282,7 +281,7 @@ def connect(self):
             try:
                 out_token = context.step(in_token)
             except spnego.exceptions.SpnegoError as err:
-                raise SMBAuthenticationError("Failed to authenticate with server: %s" % str(err.message))
+                raise SMBAuthenticationError("Failed to authenticate with server") from err
```
jborean93, line 284: Ditto for keeping the existing err message.

src/smbprotocol/exceptions.py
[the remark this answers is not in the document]
jborean93: Thanks for the info, makes sense.

src/smbprotocol/session.py
dolfinus: Ok

jborean93 (approved):
Thanks for the fixes, good to be free of some of those Python 2-isms.

Issue #250, opened 2023-11-23
Difference between SMBDirEntry._dir_info and .stat()


rikroe:
Hi,

many thanks for your library! I am exploring to use it to connect to Windows Server SMB shares.
This lead to some questions on the difference of the time objects retrieved from `scandir()` vs. `stat()`.

#### Use case
1. Listing all files in a directory 
2. Filtering new or modified files based on modified date  against a checkpoint
3. Parsing the final set of files

I'm looking for a very efficient way to do this, as sometimes we have >50k files in a single folder (we try to reduce this, but unfortunately this is not always possible).

#### Test
I started working with a smaller directory of ~150 files/folders and already see a large (order of magnitude) runtime difference, depending on the source of modified date access I am using. 

For both tries, I am using `scandir()`. It is very fast (122ms) if I extract the `last_write_time` via `SMBDirEntry._dir_info["last_write_time"]`.
If use `SMBDirEntry.stat().st_mtime`, it takes more than an order of magnitude longer (1.92s, I have also seen 2.5s).
My guess is that this will even increase as more files have to be checked, as the main issue is probably the network latency due to additional stat calls.

Speed-wise, I am leaning to just use `_dir_info`, because it is fast. However, it is a private attribute and there are also some slight differences of 1 ms (?!?) on some files/folders (see below).

#### Question
The main question would be if it is safe to use `_dir_info` to get (at least second level granularity) modification dates in an efficient way.

#### Example
(The data was only stored in a dict to make easier comparisons on single files later on. It does not have any performance impact.)

![image](https://github.com/jborean93/smbprotocol/assets/42204099/3064ac04-3ea2-4333-9e3b-45cccd74eb72)

Thank you!

jborean93:
> The main question would be if it is safe to use _dir_info to get (at least second level granularity) modification dates in an efficient way.

I don't have any plans on changing it but as you said it's not part of the public API. I can see the benefits of exposing some of the fields in the `FileIdFullDirectoryInformation` object that exposes but I would probably just make them their own read only property on the `SMBDirEntry` object or at least expose a new custom dataclass/namedtuple with the data it returns rather than expose the underlying low level API object.

Unfortunately the `stat()` call requires a bit more information than what the full directory info one returns. It could omit at least one of the query operations it requires but it's still going to be the same round trips just with a tiny bit less work than what the server needs to do.

[...]

PR #251, opened 2023-11-24
Add smb_info to a scandir result

jborean93:
Adds the property smb_info to the scandir enumerated result which contains the already retrieved metadata on the file during the scan operation. This provides a more efficient way for callers to retrieve information like the file times, sizes, attributes straight away without having to call .stat() which results in another SMB request to the server.

Fixes: https://github.com/jborean93/smbprotocol/issues/250

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 
-## 1.12.1 - TBD
+## 1.13.0 - TBD
+
+* Added the property `smb_info` on `SMBDirEntry` which returns a named tuple `SMBDirEntryInformation` containing metadata already retrieved in the `scandir` operation.
+  * This avoid having to call `stat()` to retrieve data like the file attributes or datetime fields that is already available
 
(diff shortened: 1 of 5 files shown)


rikroe:
Thank you so much!

I was hoping you'd be fine with using the data and wanted to ask to create a PR for that, but you have been blazing fast!

I tested your PR code, it is working smoothly! And even faster on large files than `ls -Al` on a cifs mount 🔥 (~4:30min against ~10min for 280k files).

rikroe (approved):
LGTM!

jborean93:
> I was hoping you'd be fine with using the data and wanted to ask to create a PR for that, but you have been blazing fast!

You just picked the right time to open the issue, I've been neglecting this library for a while now and needed to get back to it.

> I tested your PR code, it is working smoothly! And even faster on large files than ls -Al on a cifs mount 🔥 (~4:30min against ~10min for 280k files).

Thanks for testing, I'm actually surprised it's faster than a cifs mount but always happen to see that there are some benefits here :)

Issue #242, opened 2023-09-07
Unknown NtStatus error returned 'STATUS_FS_DRIVER_REQUIRED'


andrewsiemer:
I am running into the following issue when trying to copy a file from a DFS share.
```
Traceback (most recent call last):
   File "/Users/andrewsiemer/Documents/NVIDIA/asiemer_macbook/sw/main/bios/brf/Scripts/Python/Library/netutils.py", line 155, in copy_file
     with smbclient.open_file(src, mode="rb") as infile:
   File "/Users/andrewsiemer/Library/Python/3.9/lib/python/site-packages/smbclient/_os.py", line 389, in open_file
     raw_fd.open()
   File "/Users/andrewsiemer/Library/Python/3.9/lib/python/site-packages/smbclient/_io.py", line 463, in open
     transaction.commit()
   File "/Users/andrewsiemer/Library/Python/3.9/lib/python/site-packages/smbclient/_io.py", line 349, in commit
     raise failures[0]
 smbprotocol.exceptions.SMBOSError: [Error 0] [NtStatus 0xc000019c] Unknown NtStatus error returned 'STATUS_FS_DRIVER_REQUIRED': '\\share\builds\path\to\file.run'
```
The code is:
```
import smbclient.shutil

smbclient.shutil.copyfile(
    "\\share\builds\path\to\remote\file.zip",
    '\path\to\local\file.zip',
    username='adminuser',
    password='mypassword')
```

jborean93:
What is the service you are using here, is it a Windows DFS implementation/something else? I've seen a few reports for this but was never able to figure out what was happening. Typically you see `STATUS_FS_DRIVER_REQUIRED` when a DFS referral request was sent to a service that cannot handle that IOCTL code but in this case it seems like the open method. If it's possible it would be great to see the Wireshark trace for this but I can understand if the data is sensitive and cannot be shared.

andrewsiemer:
Yes, sadly this is not a share that I set up. This is the result of `smbutil statshares -a` if that's even helpful. Sorry I cannot share much more. Is there any other non-sensitive info I can share?

``` 
                              SERVER_NAME                   test
                              USER_ID                       501
                              SMB_NEGOTIATE                 SMBV_NEG_SMB1_ENABLED
                              SMB_NEGOTIATE                 SMBV_NEG_SMB2_ENABLED
                              SMB_NEGOTIATE                 SMBV_NEG_SMB3_ENABLED
[...]

PR #253, opened 2023-11-25
More advanced DFS compatibility checks

jborean93:
Avoid setting the SMB2_FLAGS_DFS_OPERATIONS flag in the SMB header if the server did not report it as supporting DFS operations during the negotiation phase. This should hopefully fix up some problems when the server errors in the case it doesn't support DFS operations.

Fixes: https://github.com/jborean93/smbprotocol/issues/242


jborean93:
This didn't help

PR #257, opened 2023-12-10
Added default timeout on connection.receive()

Robbert-Brand:
The library hang when following these steps:
 - Connect using register_connection() to a SMB server.
 - Shut down the server, using the normal shutdown procedure.
 - Call any kind of disconnect(), like reset_connection_cache()
Now the library hangs.

This is because of the connection.receive() function is called against a dead server, without a timeout.

This pull request suggests setting the default timeout of the connection.receive() function to the same value as the default timeout of the other functions.

### DIFF
--- a/src/smbprotocol/connection.py
+++ b/src/smbprotocol/connection.py
@@ -989,7 +989,7 @@ class Connection:
         """
         return self._send(messages, session_id=sid, tree_id=tid, related=related)
 
-    def receive(self, request, wait=True, timeout=None, resolve_symlinks=True):
+    def receive(self, request, wait=True, timeout=60, resolve_symlinks=True):
         """
         Polls the message buffer of the TCP connection and waits until a valid
         message is received based on the message_id passed in.
--- a/src/smbprotocol/connection.py
+++ b/src/smbprotocol/connection.py
@@ -989,7 +989,7 @@ class Connection:
         """
         return self._send(messages, session_id=sid, tree_id=tid, related=related)
 
-    def receive(self, request, wait=True, timeout=60, resolve_symlinks=True):
+    def receive(self, request, wait=True, timeout=None, resolve_symlinks=True):
         """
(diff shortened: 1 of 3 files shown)


jborean93:
Sorry it's getting to the holiday season so I haven't been able to get to this. An initial impression gives me conerns around changing the default value here as I know some tools rely on it waiting indefinitely for a response, say a pipe read. It'll require some closer investigation to see what the best way forward for this is.

Robbert-Brand:
Perhaps an alternative option would be to pass an (optional) timeout value with the disconnect and other high-level interface functions? That way, the app only hangs when the user wishes it to do so.

jborean93:
I think at least making sure `disconnect()` doesn't hang forever is a good middle ground for now. It's a pity I painted myself in the corner here with the defaults but what is done is done.

Robbert-Brand:
Hi!
I updated the pull request as you suggested. It is now possible to supply a timeout with the disconnect() functions, and the delete_session that calls the disconnect() functions defaults to the standard timeout of 60 seconds.
I also changed back the receive timeout to None, as it was originally.
Robbert

src/smbprotocol/connection.py
```diff
@@ -989,7 +989,7 @@ def send_compound(self, messages, sid, tid, related=False):
         """
         return self._send(messages, session_id=sid, tree_id=tid, related=related)
 
-    def receive(self, request, wait=True, timeout=None, resolve_symlinks=True):
+    def receive(self, request, wait=True, timeout=60, resolve_symlinks=True):
```
jborean93, line 992: Did you intend to leave this as `60` and not `None` as the default?

Robbert-Brand:
Ah, I forgot to change that one back. My apologies, now the default of receive() should be as it originally was.

jborean93 (approved):

jborean93:
Thanks for working on this, sorry for the delay in the original review.

PR #262, opened 2024-01-26
Wait for message thread to stop on disconnect.

### LINKED ISSUE #254: On disconnectt wait for msg_worker thread to end.
This should not be an issue in production.

It's just something to help with the automated tests.

For the automated tests, I have some extra checks to make sure the tests are not generating "stray" threads.

Since smbprotocol is not waiting for the `msg_thread-` to end on disconnect my cleanup checks is "flaky" and sometimes it reports that the `msg_worker-` thread was left unclosed by the test.

The `msg_worker` will enventuall close

It would be nice to have something like `self._t_worker.join(timeout=5)` at the end of `smbprotocol.connection.Connection.disconnect`

https://github.com/jborean93/smbprotocol/blob/37512ee0648ad64f98755833382fea790d9b2df6/src/smbprotocol/connection.py#L934-L949

Thanks

adiroiban:
## Scope

Fixes #254 

This is mostly to help with autoamted tests.

## Changes

On disconnect wait a bit for the thread.

If the thread is still running, disconect will still return, just like before.

I have created a fixte to reduce the duplicate code required to setup a connection and all the try/finally for cleanup.

As as drive-by I have updated the docstring for echo test and wrote another test for echo, as I was using those tests to learn about testing code.

As a drive by, I have updated the README with some info a found during initial dev.
Feel free to revert those changes.

### DIFF
--- a/src/smbprotocol/connection.py
+++ b/src/smbprotocol/connection.py
@@ -861,11 +861,11 @@ class Connection:
         log.info("Setting up transport connection")
         self.transport = Tcp(self.server_name, self.port, timeout)
         self.transport.connect()
-        t_worker = threading.Thread(
+        self._t_worker = threading.Thread(
             target=self._process_message_thread, name=f"msg_worker-{self.server_name}:{self.port}"
         )
-        t_worker.daemon = True
-        t_worker.start()
+        self._t_worker.daemon = True
+        self._t_worker.start()
 
         log.info("Starting negotiation with SMB server")
         enc_algos = preferred_encryption_algos or [
@@ -947,6 +947,7 @@ class Connection:
 
         log.info("Disconnecting transport connection")
         self.transport.close()
+        self._t_worker.join(timeout=2)
 
     def send(
         self, message, sid=None, tid=None, credit_request=None, message_id=None, async_id=None, force_signature=False
--- a/src/smbprotocol/connection.py
+++ b/src/smbprotocol/connection.py
@@ -820,6 +820,9 @@ class Connection:
         # Keep track of the message processing thread's potential traceback that it may raise.
         self._t_exc = None
 
+        # The thread that will handle message processing.
+        self._t_worker = None
+
     def connect(self, dialect=None, timeout=60, preferred_encryption_algos=None, preferred_signing_algos=None):
         """
         Will connect to the target server and negotiate the capabilities
--- a/src/smbprotocol/connection.py
+++ b/src/smbprotocol/connection.py
@@ -947,7 +947,8 @@ class Connection:
 
         log.info("Disconnecting transport connection")
         self.transport.close()
-        self._t_worker.join(timeout=2)
+        if self._t_worker:
+            self._t_worker.join(timeout=2)
 
     def send(
         self, message, sid=None, tid=None, credit_request=None, message_id=None, async_id=None, force_signature=False
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -1306,8 +1306,7 @@ class TestConnection:
         """
         connection, session = connected_session
 
-        actual = connection.echo(
-            sid=session.session_id, credit_request=65535)
+        actual = connection.echo(sid=session.session_id, credit_request=65535)
 
         assert actual == 8129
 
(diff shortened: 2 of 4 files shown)


src/smbprotocol/connection.py
```diff
@@ -947,6 +947,7 @@ def disconnect(self, close=True, timeout=None):
 
         log.info("Disconnecting transport connection")
         self.transport.close()
+        self._t_worker.join(timeout=2)
```
adiroiban, line 950: I did a short wait. mostly to help with testing.

For real world troubleshooting, where we should check and see if the thread is still running after the join and maybe log a warning ... or raise an exception.

For this PR, I wanted to keep the scope only to help with automated tests

tests/test_connection.py
```diff
@@ -1183,6 +1200,9 @@ def test_send_invalid_tree_id(self, smb_real):
             connection.disconnect()
 
     def test_connection_echo(self, smb_real):
+        """
```
adiroiban, line 1203: I have only added a docstring while trying to understand the scope/purpose of this test.

If you are happy with the new fixture, I can create a separate PR and update all the existing tests to use the fixture and remove the duplicate code.

adiroiban:
This is ready for review. Thanks!

tests/test_connection.py
```diff
@@ -34,6 +35,22 @@
...
+    SMB server and that will disconnect at the end.
+    """
+    connection = Connection(uuid.uuid4(), smb_real[2], smb_real[3])
+    connection.connect()
+    session = Session(connection, smb_real[0], smb_real[1])
+    session.connect()
+
+    yield connection, session
+
+    connection.disconnect(True)
```
jborean93, line 51: Might be best to wrap this in a try/finally

```suggestion
    try:
        yield connection, session
    finally:
        connection.disconnect(True)
```

tests/test_connection.py
jborean93: That sounds nice, don't feel obliged to do this though.

src/smbprotocol/connection.py
```diff
@@ -947,6 +950,8 @@ def disconnect(self, close=True, timeout=None):
 
         log.info("Disconnecting transport connection")
         self.transport.close()
+        if self._t_worker:
+            self._t_worker.join(timeout=2)
```
jborean93, line 954: Have you noticed this timeout being hit in real life? If it is timing out it might be worth investigating why the thread is still running.

adiroiban: I have never experienced the timeout.
This is why I didn't bother to do a check for timeout.

The thread stops... it's just that when running the automated tests, I am askinging for the list of active threads, just before the thread stop.

also, this is really a problem only during automated testing.

tests/test_connection.py
adiroiban: I am not familiar with pytest.

My understanding is that anything after yield is automatically a teardown...so finally is not needed.

https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#handling-errors-for-yield-fixture

jborean93: Ah nice to know, I always thought you had to handle the exceptions but a quick test shows that's not the case.

```python
import pytest


@pytest.fixture
def my_fixture():
    print("Fixture start")

    yield 1

    print("Fixture end")


def test_example(my_fixture):
    assert False
```

```bash
$ python -m pytest /tmp/test_example.py
```

```
====================================================================================== FAILURES ======================================================================================
____________________________________________________________________________________ test_example ____________________________________________________________________________________

my_fixture = 1

    def test_example(my_fixture):
>       assert False
E       assert False

/tmp/test_example.py:14: AssertionError
------------------------------------------------------------------------------- Captured stdout setup --------------------------------------------------------------------------------
Fixture start
------------------------------------------------------------------------------ Captured stdout teardown ------------------------------------------------------------------------------
Fixture end
```

tests/test_connection.py
```diff
@@ -1278,3 +1298,40 @@ def mockurandom(length):
...
+        connection, session = connected_session
+
+        actual = connection.echo(sid=session.session_id, credit_request=65535)
+
+        assert actual == 8129
+
+    def test_disconnect_wait_for_thread(self, connected_session):
+        """
+        When requesting more credit from the server, it will return the
+        allocated credits.
```
adiroiban, line 1316:
```suggestion
        Disconnect will continue after the message thread is stopped.
```

jborean93 (approved):

jborean93:
Thanks for working on this.

PR #285, opened 2024-08-14
Customizable receive timeout for SMB Connection

leonnis12:
This PR extends the configurability of the SMB Connection class and register_session methods to allow modifying the transport.recv timeout used for both keep-alive checks that keep Windows from closing a connection at ~16 minutes, and for checking the health of the connection.

In the usecase of my application, the SMB connection is used over an unstable network, which sometimes loses the connections between the machines. This triggers Connection timed out errors from the health check implemented in _process_message_thread, but only after 10+ minutes (the hardcoded timeout) where the application hangs. I would like to be able to configure this timeout to be smaller for use cases like this, to perform the health check more often, and be able to handle the timeout error in a reasonable ammount of time.

### Current behaviour:
- The connection is checked every 10 minutes using a SMB2_ECHO command.
- If the connection is broken due to a network problem, the connection will be stuck until the timeout passes, the ECHO command is sent and the server does not respond to it.

### Expected Behavior
- The timeout should be configurable for the cases where network errors are common, and handling them often requires waiting 10+ minutes for each such problem.

This might also be a fix for https://github.com/jborean93/smbprotocol/issues/117 (mentions a similar problem), by allowing configuring a smaller timeout value, which does not keep the application stuck for long and allows the error to be handled in my application in a timely manner. 

The PR extends the health checks added in https://github.com/jborean93/smbprotocol/pull/135 with the configurable health check interval value.

### DIFF
--- a/src/smbclient/_pool.py
+++ b/src/smbclient/_pool.py
@@ -230,7 +230,7 @@ def delete_session(server, port=445, connection_cache=None, timeout=60):
 
 
 def get_smb_tree(
-    path, username=None, password=None, port=445, encrypt=None, connection_timeout=60, connection_cache=None
+    path, username=None, password=None, port=445, encrypt=None, receive_timeout=600, connection_timeout=60, connection_cache=None,
 ):
     """
     Returns an active Tree connection and file path including the tree based on the UNC path passed in and other
@@ -244,6 +244,7 @@ def get_smb_tree(
     :param port: The port to connect with.
     :param encrypt: Whether to force encryption or not, once this has been set to True the session cannot be changed
         back to False.
+    :param receive_timeout: Override the timeout used for receiving a response from the server
     :param connection_timeout: Override the timeout used for the initial connection.
     :param connection_cache: Connection cache to be used with
     :return: The TreeConnect and file path including the tree based on the UNC path passed in.
@@ -261,6 +262,7 @@ def get_smb_tree(
         "password": password,
         "port": port,
         "encrypt": encrypt,
+        "receive_timeout": receive_timeout,
         "connection_timeout": connection_timeout,
         "connection_cache": connection_cache,
     }
@@ -307,6 +309,7 @@ def get_smb_tree(
         password=password,
         port=port,
         encrypt=encrypt,
+        receive_timeout=receive_timeout,
         connection_timeout=connection_timeout,
         connection_cache=connection_cache,
         auth_protocol=auth_protocol,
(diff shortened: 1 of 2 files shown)


alin-opswat (approved):

DragosFlorea:
This would be very helpful for my use case as well

jborean93:
Thanks for the PR.

I am somewhat weary about exposing the retry mechanism timeout as a public API as I feel like the implementation is a bit of a bandaid over the real problem rather than a good solution. I know I've been dragging my heels on trying to rework the TCP transport side where my aim was to improve this situation but I'm on the fence about accepting this type of change.

The proper solution in my mind is to figure out how to detect when the socket has been "dropped" but potentially in your situation the client still thinks its connected and thus will wait for the full timeout.

DragosFlorea:
I understand your concern... And yes the best solution is to know exactly when the connection has been dropped but a logic on the client side to cut off the connection after timeout is not bad either. If you do not want to expose the timeout as a public api, which is a low level configuration, what do you say about a environment variable marked as experimental until you have a better solution?

jborean93:
I think I can justify an environment variable here, especially if it's helping in some situations. I would have to stress it wouldn't be covered under the public API and could change/be removed in the future.

The socket stuff can be complicated, if you know the reason why the socket was closed but never reported back to the smbprotocol then I'll love to see a reproducer for it. That'll help me test out new scenarios in the future and hopefully avoid the need for the timeout altogether.

leonnis12:
I agree that the proper solution would be to detect when the socket has been "dropped", but also that the environment variable "fix" should be enough to help in the situations where this happens, until a proper solution can be found at least.

I've updated the pull request to not expose the timeout through the public API, and instead be available to be overridden through an environment variable marked as experimental from the Connection class.

Regarding reproducing the issue, simulating a network that drops packets from time to time, or has random response time spikes should be enough to trigger this issue, for further testing in the future. The following is an example from the network where I encountered the issue constantly.
pings

jborean93 (approved):

jborean93:
I'll need to sort out the CI problems on macOS separately, I believe the image version now uses the arm64 builds so the x86_64 ones  are failing without further work needed. That's not your problem to deal with so thanks for the changes and agreeing to the env var solution for now!

leonnis12:
Would it be possible to make a minor release which includes this fix, until the changes for 1.14 are ready? It would be really helpful.

jborean93:
v1.14.0 has been published on PyPI with this change https://pypi.org/project/smbprotocol/1.14.0/.

PR #294, opened 2024-10-14
Add `smbclient.liststreams` to enumerate ADS streams

### LINKED ISSUE #292: smbclient: xattr commands set Extended Attributes. Should they use Alternate Data Streams instead?
I've fallen into a bit of a rabbit hole today regarding xattrs.

There are currently two ways (connecting to a Windows SMB server, in this case my computer set to share a folder) to write extra attributes to files.

**Alternate Data Streams, which uses the [FileStreamInformation](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/f8762be6-3ab9-411e-a7d6-5cc68f70c78d) type**
```python
with smbclient.open_file(path + ":ntfs_stream", "wb") as f:
    f.write(b"Using FileStreamInformation")
```
Also seen in the [smbprotocol tests](https://github.com/jborean93/smbprotocol/blob/2bed7bc87a684251427e4bbba4f0e1d3d5e00665/tests/test_smbclient_os.py#L821-L841).

**Extended Attributes, which uses the [FileFullEaInformation](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/0eb94f48-6aac-41df-a878-79f4dcfd8989) type**
```python
smbclient.setxattr(path, "extended_attribute", "Using FileFullEaInformation")
```

The former is what `ntfs-3g` appears to map the concept of linux xattrs to ([seen here](https://github.com/tuxera/ntfs-3g/wiki/Using-Extended-Attributes)), and is viewable using commandline tools on Windows:
```cmd
D:\>dir /r test.txt
 Directory of D:\

11/10/2024  08:48 PM                12 test.txt
                                    27 test.txt:ntfs_stream:$DATA
               1 File(s)             12 bytes
               0 Dir(s)  xxx bytes free
```

The latter, far as I can tell, is only retrievable by Windows C APIs such as NtSetEaFile ([seen here](https://github.com/jschicht/EaTools)).

Anecdotally, I can also confirm that macOS maps xattr calls over mounted SMB shares to the FileStreamInformation type.

In general, it's really very hard to find information on "NTFS Extended Attributes" because most people call Alternate Data Streams "Extended attributes" when they write about them. The best summary I found was part of [an addendum to a SO question](https://superuser.com/a/1736010).

Based on what other clients and servers do, would it be reasonable to move smbclient's xattr implementation to instead use FileStreamInformation instead of FileFullEaInformation? Or did I miss something obvious!

mon:
A simple change in theory, let me know if I messed up anything with layout / documentation etc.

I haven't got a huge array of machines to test against, but it works on my test device.

Closes #292

### DIFF
--- a/src/smbclient/_os.py
+++ b/src/smbclient/_os.py
@@ -1,6 +1,8 @@
 # Copyright: (c) 2019, Jordan Borean (@jborean93) <jborean93@gmail.com>
 # MIT License (see LICENSE or https://opensource.org/licenses/MIT)
 
+from __future__ import annotations
+
 import collections
 import datetime
 import errno
@@ -37,6 +39,7 @@ from smbprotocol.file_info import (
     FileLinkInformation,
     FileRenameInformation,
     FileStandardInformation,
+    FileStreamInformation,
 )
 from smbprotocol.header import NtStatus
 from smbprotocol.ioctl import (
@@ -100,6 +103,12 @@ SMBStatVolumeResult = collections.namedtuple(
         "actual_available_size",
     ],
 )
+
+
+class SMBFileStreamInformation(t.NamedTuple):
+    name: str
+    size: int
+    allocation_size: int
 
 
 def is_remote_path(path: str) -> bool:
@@ -1017,6 +1026,43 @@ def setxattr(path, attribute, value, flags=0, follow_symlinks=True, **kwargs):
         ea_info["ea_name"] = b_attribute
         ea_info["ea_value"] = b_value
         set_info(transaction, ea_info)
+
+
+def liststreams(path: str, follow_symlinks=True, **kwargs: t.Any) -> list[SMBFileStreamInformation]:
+    """
+    Return a list of the alternative data streams on a path. Listed streams can
+    be opened by appending their name to the original path. An example call for
+    a file with a single extra stream may return:
+
+    ```
+    [
+        SMBFileStreamInformation(name=':extra_stream:$DATA', size=8, allocation_size=8),
+        SMBFileStreamInformation(name='::$DATA', size=103472, allocation_size=131072),
+    ]
+    ```
+
+    :param path: The full UNC path to the file to get the list of streams for.
+    :param follow_symlinks: Whether to follow the symlink at path if encountered.
+    :param kwargs: Common SMB Session arguments for smbclient.
+    :return: List of streams on the file with each entry being a string.
+    """
+
+    raw = SMBRawIO(
+        path,
+        desired_access=FilePipePrinterAccessMask.FILE_READ_ATTRIBUTES,
+        create_options=0 if follow_symlinks else CreateOptions.FILE_OPEN_REPARSE_POINT,
+        **kwargs,
+    )
+
+    with SMBFileTransaction(raw) as transaction:
+        query_info(transaction, FileStreamInformation, output_buffer_length=MAX_PAYLOAD_SIZE)
+
+    return [
+        SMBFileStreamInformation(
+            s["stream_name"].get_value(), s["stream_size"].get_value(), s["stream_allocation_size"].get_value()
+        )
+        for s in transaction.results[0]
+    ]
 
 
 def _delete(raw_type, path, **kwargs):
--- a/src/smbclient/__init__.py
+++ b/src/smbclient/__init__.py
@@ -14,6 +14,7 @@ from smbclient._os import (
     getxattr,
     link,
     listdir,
+    liststreams,
     listxattr,
     lstat,
     makedirs,
--- a/tests/test_smbclient_os.py
+++ b/tests/test_smbclient_os.py
@@ -834,10 +834,7 @@ def test_open_file_with_ads(smb_share):
 
     assert smbclient.listdir(smb_share) == ["file.txt"]
 
-    with smbclient.open_file(filename, buffering=0, mode="rb") as fd, SMBFileTransaction(fd) as trans:
-        query_info(trans, FileStreamInformation, output_buffer_length=1024)
-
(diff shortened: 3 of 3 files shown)


src/smbclient/_os.py
```diff
@@ -1019,6 +1020,36 @@ def setxattr(path, attribute, value, flags=0, follow_symlinks=True, **kwargs):
         set_info(transaction, ea_info)
 
 
+def liststreams(path, follow_symlinks=True, **kwargs) -> list[str]:
```
jborean93, line 1023: Using `list[str]` will require the future import at the top of the file for older Python versions.

```python
from __future__ import annotations
```

While we haven't typed any of the other functions yet it would be great to start doing it for new additions

```suggestion
def liststreams(
    path: str,
    follow_symlinks: bool = True,
    **kwargs: typing.Any,
) -> list[str]:
```

src/smbclient/_os.py
```diff
@@ -1019,6 +1020,36 @@ def setxattr(path, attribute, value, flags=0, follow_symlinks=True, **kwargs):
...
+    :param follow_symlinks: Whether to follow the symlink at path if encountered.
+    :param kwargs: Common SMB Session arguments for smbclient.
+    :return: List of streams on the file with each entry being a string.
+    """
+
+    raw = SMBRawIO(
+        path,
+        mode="r",
+        share_access="r",
+        create_options=0 if follow_symlinks else CreateOptions.FILE_OPEN_REPARSE_POINT,
```
jborean93, line 1043: Can a reparse point have an ADS, does it make sense to not follow symlinks if it cannot?

src/smbclient/_os.py
```diff
@@ -1019,6 +1020,36 @@ def setxattr(path, attribute, value, flags=0, follow_symlinks=True, **kwargs):
...
+        **kwargs,
+    )
+
+    with SMBFileTransaction(raw) as transaction:
+        query_info(transaction, FileStreamInformation, output_buffer_length=MAX_PAYLOAD_SIZE)
+
+    return [s["stream_name"].get_value() for s in transaction.results[0]]
```
jborean93, line 1050: Should we return a named tuple of `name`, `size`, and `allocation_size` instead of just a string to make use of the full `FileStreamInformation` result.

src/smbclient/_os.py
```diff
@@ -1019,6 +1020,36 @@ def setxattr(path, attribute, value, flags=0, follow_symlinks=True, **kwargs):
...
+    [":extra_stream:$DATA", "::$DATA"]
+    ```
+
+    :param path: The full UNC path to the file to get the list of streams for.
+    :param follow_symlinks: Whether to follow the symlink at path if encountered.
+    :param kwargs: Common SMB Session arguments for smbclient.
+    :return: List of streams on the file with each entry being a string.
+    """
+
+    raw = SMBRawIO(
```
jborean93, line 1039: Might need some further investigation to see what is involved here when trying to enumerate the streams on a directory. I'll play around with it a bit and get back to you on that.

jborean93 (changes requested):
Thanks for the PR, have added some comments.

mon:
I find it interesting that the tests failed on the pre-3.9 lowercase `list`, when it's used elsewhere, but I'll fix that

src/smbclient/_os.py
mon: I don't know, but I suspect it's impossible. I'll remove this arg

src/smbclient/_os.py
mon: I'm also midway through adding proper inferred types for `open_file`, though the buffering needing a "nonzero" case is causing me grief. Hope to finish shortly.

src/smbclient/_os.py
```diff
@@ -1019,6 +1022,48 @@ def setxattr(path, attribute, value, flags=0, follow_symlinks=True, **kwargs):
...
+
+    :param path: The full UNC path to the file to get the list of streams for.
+    :param kwargs: Common SMB Session arguments for smbclient.
+    :return: List of streams on the file with each entry being a string.
+    """
+
+    raw = SMBRawIO(
+        path,
+        mode="r",
+        share_access="r",
```
jborean93, line 1052: By only requesting `FILE_READ_ATTRIBUTES` we can open a handle to a file that is locked even to exclusive reads.

```suggestion
        desired_access=FilePipePrinterAccessMask.FILE_READ_ATTRIBUTES,
```

src/smbclient/_os.py
jborean93: I was wrong you can, the unamed `$DATA` stream is empty but it can certainly be added alongside others. Best to keep this add still in there.

src/smbclient/_os.py
jborean93: Looks like it just works so no extra action here needed.

src/smbclient/_os.py
```diff
@@ -1019,6 +1022,48 @@ def setxattr(path, attribute, value, flags=0, follow_symlinks=True, **kwargs):
         set_info(transaction, ea_info)
 
 
+class SMBFileStreamInformation(t.NamedTuple):
```
jborean93, line 1025: Just a minor nitpick could you move this to the top alongside the other tuples please

mon: Oh, I didn't even see them there as they're using `collections.namedtuple`. Worth upgrading them to classes now? Or just leave them alone for now

jborean93: I think just leave them alone for now. Can always change to the typing variant at a future point in time.

jborean93 (approved):

jborean93:
Thanks for working on this!

PR #295, opened 2024-10-15
Add some type annotations, most notably to `smbclient.open_file`

mon:
I won't lie, overloads are super verbose. But I worked off the ones for the stdlib `open()`, so it's "known good" at least.
I tested this by looking through all the test_smbclient_os opens and making sure the correct type was inferred.

I also added some extras so you can get typed connection details from an fd. It still kinda sucks (because something like the buffer sizes are None until `connect` is called) but at least it has intellisense.

Unfortunate API break: I couldn't work out a way to set the file/dir/pipe arg to specific values without forcing kwargs with * or moving the arg. I chose the latter, interested to hear your thoughts.

### DIFF
--- a/src/smbclient/_os.py
+++ b/src/smbclient/_os.py
@@ -1,6 +1,8 @@
 # Copyright: (c) 2019, Jordan Borean (@jborean93) <jborean93@gmail.com>
 # MIT License (see LICENSE or https://opensource.org/licenses/MIT)
 
+from __future__ import annotations
+
 import collections
 import datetime
 import errno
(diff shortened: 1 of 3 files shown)


src/smbclient/_os.py
```diff
@@ -306,17 +306,208 @@ def makedirs(path, exist_ok=False, **kwargs):
...
+    "b+x",
+    "+bx",
+]
+OpenBinaryModeWriting: t.TypeAlias = t.Literal["wb", "bw", "ab", "ba", "xb", "bx"]
+OpenBinaryModeReading: t.TypeAlias = t.Literal["rb", "br"]
+OpenBinaryMode: t.TypeAlias = OpenBinaryModeUpdating | OpenBinaryModeReading | OpenBinaryModeWriting
+FileType: t.TypeAlias = t.Literal["file", "dir", "pipe"]
+
+
+class SMBBufferedRandom(io.BufferedRandom):
```
jborean93, line 379: I think we should hold off on explicitly documenting the `raw` value as our `SMBRawIO` type (and these subsequence `SMBBuffered*` types and just document the return values as the `io.Buffered*` types for now.

src/smbclient/_os.py
```diff
@@ -306,17 +306,208 @@ def makedirs(path, exist_ok=False, **kwargs):
...
+    file_attributes=None,
+    **kwargs,
+) -> SMBBufferedWriter: ...
+
+
 def open_file(
     path,
     mode="r",
     buffering=-1,
+    file_type: t.Literal["file", "dir", "pipe"] = "file",
```
jborean93, line 504: So why does this need to be defined before the others? Is it because the overload only works based on the argument order?

src/smbclient/_os.py
mon: Ah, ok - I specifically added this because the tests access the raw FDs connection info.

Getting rid of this also lets me get rid of the tree/connection types in the other files causing the circular imports (or I can leave them in and gate with TYPE_CHECKING)

src/smbclient/_os.py
jborean93: I think let's just ignore the unbuffered file types and we can investigate a better API in the future without hopefully exposing a breaking change.

src/smbclient/_os.py
jborean93: Sounds like a win win :)

mon: I'll keep some of them (auth type, for example) but remove anything circular

mon:
Hold up on a merge just for a moment, I'm getting a runtime error on py3.10 that the tests haven't caught, making sure I didn't break anything

mon:
Alright, good to go once the tests finish running. Sorry for the churn!

jborean93 (approved):

jborean93:
Thanks for working on this

mon:
Thanks for the fast reviews, and having the most pleasant SMB library in the ecosystem :D

PR #309, opened 2025-01-27
Implement SMB2 LOCK operation

xhernandez:
This PR adds support for locking and unlocking byte ranges in files.

### DIFF
--- a/src/smbprotocol/open.py
+++ b/src/smbprotocol/open.py
@@ -257,6 +257,19 @@ class WriteFlags:
     SMB2_WRITEFLAG_WRITE_UNBUFFERED = 0x00000002
 
 
+class LockFlags:
+    """
+    [MS-SMB2] v53.0 2017-09-15
+
+    2.2.26.1 SMB2 LOCK Request Flags
+    """
+
+    SMB2_LOCKFLAG_SHARED_LOCK = 0x1
+    SMB2_LOCKFLAG_EXCLUSIVE_LOCK = 0x2
+    SMB2_LOCKFLAG_UNLOCK = 0x4
+    SMB2_LOCKFLAG_FAIL_IMMEDIATELY = 0x10
+
+
 class QueryDirectoryFlags:
     """
     [MS-SMB2] v53.0 2017-09-15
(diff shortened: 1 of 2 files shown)


src/smbprotocol/open.py
```diff
@@ -1523,3 +1602,43 @@ def _close_response(self, request):
             self.end_of_file = c_resp["end_of_file"].get_value()
             self.file_attributes = c_resp["file_attributes"].get_value()
         return c_resp
+
+    def lock(self, locks, lsn=0, lsi=0, wait=True, send=True):
```
jborean93, line 1606: Do you have a particular reason for exposing the sequence and index numbers right now? The protocol docs somewhat indicate they are used for some global state that I don't fully understand and am wondering if it is just best to keep it at 0 and expose it when requested by someone.

https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/06d42500-2ead-4659-8af2-86dcaec5286e

src/smbprotocol/open.py
```diff
@@ -1523,3 +1602,43 @@ def _close_response(self, request):
...
+            message to the caller and the unpack function
+        :return: SMB2LockResponse message received from the server
+        """
+
+        lock = SMB2LockRequest()
+        lock["file_id"] = self.file_id
+        lock["locks"] = locks
+
+        if self.connection.dialect > Dialects.SMB_2_0_2:
+            lock["lock_sequence"] = (lsn << 28) + lsi
```
jborean93, line 1631: If we do decide to expose this, we should ensure that `lsn` fits into the 28 bits and `lsi` isn't more than 4 bits. If they are then we should raise a `ValueError`.

jborean93 (commented):
I believe the CI failures is because I didn't bump the version so it is still pulling in the version from PyPI. I've just merged https://github.com/jborean93/smbprotocol/pull/310 which should solve that. You probably need to rebase to get those latest changes sorry.

src/smbprotocol/open.py
xhernandez: I exposed it because it's something that clients may use for dialects > 2.0.2. I don't fully understand it either, but I wanted to be able to reproduce any sequence of operations that a client may send to a server (the reason I implemented the LOCK operation was to reproduce an issue we detected with Samba, though it didn't use lsn/lsi).

I think it doesn't hurt to have them, and they may become useful in some cases. The protocol has them, so I don't see why we should hide them. In normal cases they won't be used, so the default value of 0 will be applied.

src/smbprotocol/open.py
xhernandez: Sure. I'll add those checks.

jborean93 (approved):

src/smbprotocol/open.py
jborean93: Fair enough, I didn't want to paint us into a corner but it's probably not going to be a problem. We can always deprecate things if we want to take away the ability to set these if we ever find out it should be calculated internally somehow but I doubt that would happen.

jborean93:
Thanks for implementing this feature!

PR #311, opened 2025-02-12
Fix TypeError: smbclient._io.SMBDirectoryIO() got multiple values for keyword argument 'share_access'

andrewsiemer:
Switched 'share_access' to default kwarg in `scandir()` to avoid multiple values error.

There is probably a reason for this error but wanting to start a thread because this solution unblocks me currently from using `share_access` with high-level functions.


jborean93:
Is there a reason why you are setting a custom `share_access` on the `scandir` method? The default is `rwd` which is all the share access so anything else is either the same or more limited in what it can open. Are you finding that there are problems with it being too open?

PR #317, opened 2025-04-08
Allow floating point times for smbclient.utime

mon:
The doc for utime:
> If times is not None, it must be a 2-tuple of the form (atime, mtime) where each member is an int or float expressing seconds.

However, providing floats throws:
```
TypeError: Cannot parse value for field last_access_time of type float to an int
```

This fixes that and adds a unit test to verify.


jborean93 (approved):
Looks great, thanks for adding the test as well!

PR #322, opened 2025-07-20
Sphinix based auto docs generator. 

srbhp:
Hi, 
Thanks a lot for your project. 
I think it will be great to have a sphinx
based docs generator for this project. 
There are still many things missing with 
this PR and would love to hear your feedback.
May you please have look at this PR?Thanks.

How to build docs: 
- run `make html` in the `docs` folder
- to browse the doc  open `build/html/index.html` in your browser.

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,0 +1,2 @@
+build/
+source/_autosummary/*
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
+#
+
+# You can set these variables from the command line, and also
+# from the environment for the first two.
+SPHINXOPTS    ?=
+SPHINXBUILD   ?= sphinx-build
+SOURCEDIR     = source
+BUILDDIR      = build
+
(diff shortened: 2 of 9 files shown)


jborean93:
Hi

I appreciate the time and effort you've put into the PR but I'm weary of adding support for generating docs. There are a few reasons for this

+ It's hard to maintain dep versions, especially around Sphinx and keeping the build working
+ Not all the functions we have are currently documented, or documented well
+ I personally prefer more work done to improve the type annotation for the various APIs instead as that's far more useful for people writing against this library

I've got a few projects that use Sphinx but I've never really been too happy with how the output docs look and also with the added work needed to ensure the docs stay buildable.

srbhp:
Thanks for you reply. 
- Yes dep management is a problem I undestand. But if you lock packages(with `uv` or other package manager) with to a particular version 
 then thing should ideally work everytime.
- We can improve this as we go 
- I do agree with this point but currently users dont know which functions does what unless they browse the source code. 
Anyway i understand your concerns. Once again thanks a lot for maintaining this project.

PR #323, opened 2025-08-05
Handle ECONNABORTED and avoid joining the current thread

avivbrg:
In windows the socket might also be terminated with `errno.ECONNABORTED`, and since `disconnect` can be called from the message thread we need to avoid `join` in that case


jborean93:
I do mean to get to this but it's a very sensitive part of the code. I'm hoping to look through this sometime in the next two weeks.

PR #327, opened 2025-10-08
Generalize to Callables Instead of Strictly Lambdas

Jayson-Fong:
For fields that accept a lambda, generalize it to accept a callable in general.

This allows some code to be simplified for programs using the package. For example:

```python
IntField(size=2, default=lambda s: len(s))
```

Can be simplified into:

```python
IntField(size=2, default=len)
```

Also updated the CI to exclude Windows 2019 as it is no longer supported (see: [actions/runner-images#12045](https://github.com/actions/runner-images/issues/12045))


jborean93 (approved):
The changes look good, just needs a rebase now that the CI has been updated.

jborean93:
Thanks for the update, changes look good.

PR #344, opened 2026-04-03
Handle DFS path resolution during directory listing

sunilkata-lb:
## Summary

- Handle `STATUS_PATH_NOT_COVERED` (and Samba equivalents `ObjectNameNotFound`, `ObjectPathNotFound`) during `query_directory()` in `SMBDirectoryIO`, enabling DFS directory listing across servers
- Close the old file handle before reassigning to a new DFS target to prevent resource leaks
- Track attempted DFS paths to prevent infinite redirect loops

## Context

The existing DFS path resolution works for file open/transaction operations (`SMBFileTransaction.commit()`), but `query_directory()` did not handle DFS redirects. When listing a directory on a DFS namespace that crosses server boundaries, the server returns `STATUS_PATH_NOT_COVERED` and the client needs to follow the DFS referral.

This mirrors the existing DFS handling pattern in `SMBFileTransaction.commit()` (line 298-330).

## Test plan

- Tested manually against a DFS namespace with links pointing to different shares
- Verified that `attempted_dfs_paths` correctly prevents infinite redirect loops

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/smbclient/_io.py
+++ b/src/smbclient/_io.py
@@ -159,6 +159,40 @@ def _resolve_dfs(raw_io):
         yield Open(tree, fd_path)
 
 
+def _next_dfs_referral(raw_io, attempted_paths):
+    """
+    Walk DFS referrals for ``raw_io`` and return the next candidate ``Open`` whose
+    (share_name, file_name) hasn't been attempted yet. Returns ``None`` if all
+    referrals are exhausted, letting the caller re-raise the original DFS error.
+
+    Used by both :class:`SMBFileTransaction` (open/CreateRequest path) and
+    :meth:`SMBDirectoryIO.query_directory` (already-open enumeration path) so
+    they share a single source of truth for DFS retry selection. Each caller
+    is still responsible for swapping the underlying fd after a candidate is
+    returned because that step differs (transaction mutates ``_actions`` while
+    the directory IO closes/reopens the fd).
+
+    :param raw_io: The SMBRawIO whose path returned STATUS_PATH_NOT_COVERED or
+        a generic name/path-not-found that's actually DFS.
+    :param attempted_paths: Mutable set used to dedupe referral targets across
+        multiple retries within a single operation. See
+        https://github.com/jborean93/smbprotocol/issues/228.
+    :return: A new ``Open`` instance for the caller to use, or ``None``.
+    """
+    for smb_open in _resolve_dfs(raw_io):
+        if smb_open.tree_connect.share_name == raw_io.fd.tree_connect.share_name:
+            continue
+
+        tested_path = f"{smb_open.tree_connect.share_name}{smb_open.file_name}".lower()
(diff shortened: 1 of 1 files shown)


src/smbclient/_io.py
```diff
@@ -615,11 +615,36 @@ class SMBDirectoryIO(SMBRawIO):
 
     def query_directory(self, pattern, info_class):
         query_flags = QueryDirectoryFlags.SMB2_RESTART_SCANS
+        attempted_dfs_paths = set()
         while True:
             try:
                 entries = self.fd.query_directory(pattern, info_class, flags=query_flags)
             except NoMoreFiles:
                 break
+            except (PathNotCovered, ObjectNameNotFound, ObjectPathNotFound):
```
jborean93, line 624: I see this is mostly a copy of the logic in the `SMBFileTransaction.commit()` method, is it at all possible to wrap the `query_directory` file in an internal transaction so it just shares the same codebase and logic? Otherwise if that's not possible maybe we need to try and strip out the logic into a common function so both of them can just re-use the same method rather than duplicating it in both places.

I'm also not 100% sure when these exceptions would fire in a DFS path, e.g. whether it's when it tries to enumerate a specific directory entry that we need to handle individually or whether it's the entire `query_directory` SMB call.

jborean93 (commented):
Thanks for working on this I've added a comment inline after a brief look over. I really would also like to see an integration test to handle this specific feature to ensure that it does actually work and won't regress again in the future.

src/smbclient/_io.py
sunilkata-lb: Apologies for the long silence on this.

Wrapping query_directory in a transaction doesn't fit cleanly: it's a streaming generator that issues one SMB2_QUERY_DIRECTORY per chunk, while a transaction is single compound send / single receive —you'd lose streaming or have to commit per chunk, and the transaction's "first action must be CreateRequest" assumption doesn't apply here.

Pushed the shared-helper alternative instead — extracted the candidate-selection part of the loop (_resolve_dfs walk + same-share skip + attempted_paths dedupe) into _next_dfs_referral(raw_io, attempted_paths). Each caller still does its own fd swap because the steps genuinely differ — commit() mutates _actions, query_directory() closes and reopens the fd. Behavior unchanged.

On when the exception fires: query_directory is one SMB2 request per call returning one batch (or an error), so the exception applies to the whole call, not per entry. DFS resolution normally happens at open time, so this is unusual on Windows servers — we hit it against a Samba/NetApp build that re-resolves on enumerate after a successful open. Happy to add a comment in source making that origin explicit if you'd like.

jborean93 (approved):

jborean93:
I tried creating an integration test but I cannot seem to replicate it locally using the test Samba server. The changes look good though just a pity we can't test it in CI.

PR #345, opened 2026-04-16
Add target_ip parameter to connection to avoid relying on internal DNS resolution

laxa:
The connect method of smbprotocol only supports server-name, which is not permissive enough.
Indeed, when using Kerberos, in certain network context (for example, relying on SOCKS connection), the DNS resolution might not work properly.
Thus, this PR adds a parameter to specify a target ip address to avoid relying on a DNS resolution in such a case.

### DIFF
--- a/src/smbprotocol/session.py
+++ b/src/smbprotocol/session.py
@@ -3,7 +3,6 @@
 
 import hashlib
 import logging
-import random
 from collections import OrderedDict
 from typing import Literal, Optional
 
@@ -177,6 +176,7 @@ class Session:
         username: Optional[str] = None,
         password: Optional[str] = None,
         require_encryption=True,
+        hostname_override=None,
         auth_protocol: Literal["negotiate", "ntlm", "kerberos"] = "negotiate",
     ):
         """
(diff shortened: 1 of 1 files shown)


laxa:
Well, it seems like this modification can not be "easily handled" in the unit tests without adding the target_ip_addresses in all the configuration of the CI/CD, let me know if I should try to modify this or if you can handle it.

jborean93:
> The connect method of smbprotocol only supports server-name, which is not permissive enough.

It certainly does support IPs, it sounds like you are trying to connect to an IP address but still wish to specify the hostname for the Kerberos SPN lookup? If so I would probably recommend we keep the `Connection` class as is but add a new kwarg to Session that can specify the hostname portion for the authentication context and change this line to `hostname=self._hostname_override or self.connection.server_name`

https://github.com/jborean93/smbprotocol/blob/c219a0f0705488baa54632378be5dc3f0e25c6f6/src/smbprotocol/session.py#L277

This way the connection logic stays focused solely on the connection itself and how to talk to the host whereas the authentication/Kerberos specific logic you want to apply when connecting to a hostname is overridden only where it actually matters.

laxa:
Sorry for not explaining clearly what I meant. But you guessed it, the goal is to specify the hostname in a SPN lookup context while not relying on DNS to perform the resolution.
I have removed the original code modifying the `Connection` class to implement the logic within the `Session` class.
I also removed the `random` import which was not used anymore.

jborean93 (approved):

jborean93:
You're all good, I just wanted to clarify things and make sure I understood what you were saying. Changes look good, will merge once CI is green.

PR #351, opened 2026-05-08
Improve recv-path diagnostics

henrik-nil-acc:
A bare TimeoutError from Tcp._recv carries no context about which recv fired or how many bytes had been consumed. The worker thread can die silently on any other Exception, leaving _t_exc to surface only on the next app call. Peer-aborted sockets (ESHUTDOWN, ECONNRESET, ECONNABORTED) are mapped to clean EOF with no log.


jborean93 (approved):
Looks great, appreciate you improving this.

PR #353, opened 2026-05-20
Stop wire close from SMBRawIO finalizer

### LINKED ISSUE #352: Concurrent  rmtree  deadlocks the SMB connection on garbage collection
## Summary

`smbclient.shutil.rmtree` and any other code that drops
`SMBRawIO` references without explicit `.close()` can permanently
hang the SMB connection worker thread. After the hang every
subsequent SMB operation in the process blocks forever.

## Root cause

`SMBRawIO` inherits `io.IOBase.__del__`
(https://github.com/python/cpython/blob/v3.12.3/Lib/_pyio.py#L407),
which calls `self.close()`. The `close()` override on `SMBRawIO`
issues an SMB2 CLOSE on the wire and waits on
`Connection.receive()`.
Replies are delivered exclusively by the worker thread, so when
cyclic GC schedules the finalizer on the worker the wait
self-deadlocks: the worker is waiting on an event only it can
set.

Python's data model documents the underlying hazard: `__del__`
"can be invoked when arbitrary code is being executed, including
from any arbitrary thread"
(https://docs.python.org/3/reference/datamodel.html#object.__del__).

Cyclic garbage is easy to generate. Concurrent `rmtree` workers
race the same paths and raise tracebacks. Each traceback creates
a frame -> traceback -> frame cycle. Cyclic GC fires on whichever
thread crosses the allocation threshold, sometimes the worker.

## Reproducer

12 spawn-mode processes (each its own connection) running
`rmtree(ignore_errors=True)` on the same share. Tightened GC
thresholds. Tree: 20 directories x 50 files.

```python
import gc, multiprocessing as mp
from smbclient import register_session
from smbclient.shutil import rmtree

def worker():
    gc.set_threshold(1, 1, 1)
    register_session("server", username="...", password="...")
    rmtree(r"\\server\share\target", ignore_errors=True)

if __name__ == "__main__":
    ctx = mp.get_context("spawn")
    procs = [ctx.Process(target=worker) for _ in range(12)]
    for p in procs: p.start()
    for p in procs: p.join(timeout=45)
    hung = [p for p in procs if p.is_alive()]
    print(f"hung: {len(hung)}/12")
```

Observed: 11/12 processes still alive after 45s.

## Stack signature

From a hung worker via `faulthandler.dump_traceback_later`. Line
numbers drift across `origin/master`. Function names and call
chain are stable.

```
Thread 0x... (most recent call first):
  File "threading.py" in wait
  File "smbprotocol/connection.py" in receive
  File "smbprotocol/open.py" in _close_response
  File "smbprotocol/open.py" in close
  File "smbclient/_io.py" in close
  File "smbprotocol/structure.py" in __init__
  File "smbprotocol/structure.py" in __init__
  File "smbprotocol/header.py" in __init__
  File "smbprotocol/connection.py" in _process_message_thread
  File "threading.py" in run
```

## Environment

- Python 3.12.3
- smbprotocol `origin/master` 1.17.0-dev
- Server: Samba 4.19.5 (Ubuntu package `samba 2:4.19.5+dfsg-4ubuntu9.4`)

---

Diagnosis assisted by Claude Code. Reproduction verified by humans.

henrik-nil-acc:
Fixes #352.

io.IOBase.__del__ would call self.close() during GC, which
issues SMB I/O and waits on Connection.receive(). On the
worker thread that wait self-deadlocks. Override
SMBRawIO.__del__ to skip the wire close and emit
ResourceWarning instead, mirroring asyncio.BaseEventLoop
and subprocess.Popen, which warn rather than perform
cleanup that would be unsafe in the finalizer's thread.

Add a Connection.receive guard that raises SMBException if
called from the worker thread, so the same mistake fails
loudly elsewhere.

The warning fires only when an SMBRawIO is dropped without
being closed. The usual idioms (with open_file(...) as f:
or explicit f.close()) prevent it; the smbclient.shutil
helpers manage their own handles.

### DIFF
--- a/src/smbclient/_io.py
+++ b/src/smbclient/_io.py
@@ -3,6 +3,7 @@
 
 import io
 import logging
+import warnings
 
 from smbclient._pool import ClientConfig, dfs_request, get_smb_tree
 from smbprotocol import MAX_PAYLOAD_SIZE
@@ -423,6 +424,17 @@ class SMBRawIO(io.RawIOBase):
         }.get(self.FILE_TYPE, 0)
 
         super().__init__()
+
+    def __del__(self):
+        # io.IOBase.__del__ would call self.close(), which self-deadlocks if GC
+        # runs it on the SMB worker thread.
+        try:
+            closed = self.closed
+        except AttributeError:
+            return
+        if closed:
+            return
+        warnings.warn(f"unclosed SMB handle {self._name!r}", ResourceWarning, source=self)
 
     def __enter__(self):
         self.open()
--- a/src/smbprotocol/connection.py
+++ b/src/smbprotocol/connection.py
@@ -1033,6 +1033,12 @@ class Connection:
         :param resolve_symlinks: Set to automatically resolve symlinks in the path when opening a file or directory.
         :return: SMB2HeaderResponse of the received message
         """
+        # Only the worker thread delivers responses, so calling receive() from it would wait
+        # on an event only it can set.
+        worker = self._t_worker
+        if worker is not None and threading.get_ident() == worker.ident:
+            raise SMBException("Connection.receive() called from the SMB worker thread, would self-deadlock")
+
         # Make sure the receiver is still active, if not this raises an exception.
         self._check_worker_running()
 
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -1115,6 +1115,16 @@ class TestConnection:
         with pytest.raises(SMBConnectionClosed):
             connection.receive(None)
 
+    def test_receive_from_worker_thread_raises(self, mocker):
+        # receive() on the worker thread would wait for an event only the worker can set.
+        connection = Connection(uuid.uuid4(), "server", 445, True)
+        fake_worker = mocker.MagicMock()
+        fake_worker.ident = threading.get_ident()
+        connection._t_worker = fake_worker
+
+        with pytest.raises(SMBException, match="self-deadlock"):
+            connection.receive(None)
+
     def test_verify_fail_no_session(self, smb_real):
         connection = Connection(uuid.uuid4(), smb_real[2], smb_real[3], True)
         connection.connect()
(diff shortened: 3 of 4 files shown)


src/smbclient/_io.py
```diff
@@ -424,6 +425,18 @@ def __init__(
...
+        # io.IOBase.__del__ would call self.close(), which does SMB I/O and waits
+        # on Connection.receive(). On the worker thread that wait self-deadlocks.
+        # _warn captures warnings.warn so it survives interpreter teardown.
+        try:
+            closed = self.closed
+        except AttributeError:
+            return
+        if closed:
+            return
+        _warn(f"unclosed SMB handle {self._name!r}", ResourceWarning, source=self)
```
jborean93: `_warn` doesn't seem to be defined so it sounds like this would fail? Is the goal for this to be a no-op with the warning added?

henrik-nil-acc: _warn is defined at line 428.
Yes, it's supposed to be warning and no-op. 
This means a resource leak, not good but better than deadlock.
There are multiple locations that needs fixes in the error paths to avoid that and I plan follow-up PR's once this is merged.

henrik-nil-acc:
Anything I can do to help to git this merged? The issue has a reproducer, it will trigger a bunch of other issues that I have fixes for once this is merged.

src/smbclient/_io.py
jborean93: Is there a reason why you used the pattern to define the function as a kwarg default? Why not just do `warnings.warn(...)` instead?

jborean93:
Sorry I've been away for a bit so have had limited GitHub interaction. Just added another reply to the change.

src/smbclient/_io.py
henrik-nil-acc: The pattern was copied from [`subprocess.Popen.__del__`](https://github.com/python/cpython/blob/main/Lib/subprocess.py#L1300). I guess there is a theoretical chance the "warnings" is teared down at the time of calling __del__ if not holding a reference. But I can remove that part if you want me to? I have not seen it happen.

jborean93: I think so, this is certainly a different pattern than what I was expecting and honestly prefer it to just hard break if such an event happened. We can always go back if we find it's an actual problem.

henrik-nil-acc: Done

jborean93 (approved):

PR #356, opened 2026-06-09
Fix signing requirement against signing required

laxa:
When using `smbprotocol`, with `require_signing=False` on the `Connection` class, against a computer requiring signing (a Domain Controller for example), the client fails to update its `client_security_mode` as stated in the [specification](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/5606ad47-5ee0-437a-817e-70c366052962) section 3.2.4.2.2.2.

image

EDIT: force pushed to update test case that failed with the initial PR


jborean93 (approved):
Thanks for the fix

PR #358, opened 2026-06-23
Treat STATUS_NO_SUCH_FILE as end of enumeration

henrik-nil-acc:
MS-SMB2 3.3.5.18 permits servers to terminate QUERY_DIRECTORY
with either STATUS_NO_MORE_FILES or STATUS_NO_SUCH_FILE. Samba
returns the latter on a non-matching pattern, but existing code
caught only the first. The escaping NoSuchFile is not an
OSError subclass, so it slips past os-style except clauses and
rmtree's ignore_errors.

### DIFF
--- a/src/smbclient/_io.py
+++ b/src/smbclient/_io.py
@@ -9,6 +9,7 @@ from smbclient._pool import ClientConfig, dfs_request, get_smb_tree
 from smbprotocol import MAX_PAYLOAD_SIZE
 from smbprotocol.exceptions import (
     NoMoreFiles,
+    NoSuchFile,
     ObjectNameNotFound,
     ObjectPathNotFound,
     PathNotCovered,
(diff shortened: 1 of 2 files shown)


jborean93:
Does this actually happen in real life with an SMB implementation I don't know of? I tried it against a Windows host and it didn't report this error just no items to yield. The fact that the test has to have a mock to represent this type of error also tells me that maybe Samba doesn't do this. While the comment indicate that MS-SMB2 shows `STATUS_NO_SUCH_FILE` can be returned it doesn't say why and if I can't see it through Windows then I'm wondering if it is returned for another reason that we don't want to ignore.

henrik-nil-acc:
> Does this actually happen in real life with an SMB implementation I don't know of? I tried it against a Windows host and it didn't report this error just no items to yield. The fact that the test has to have a mock to represent this type of error also tells me that maybe Samba doesn't do this. While the comment indicate that MS-SMB2 shows `STATUS_NO_SUCH_FILE` can be returned it doesn't say why and if I can't see it through Windows then I'm wondering if it is returned for another reason that we don't want to ignore.

It happens with samba. New integration tests should be failing before the change. Also enabling debug logs should make it visible.

jborean93:
> It happens with samba

Ah I see, it was the non-matching pattern that did the trick. I was just doing `scandir` on an empty directory with the default pattern `*`. Looks like MS' SMB server does the same thing with a pattern that doesn't match.

Would it be possible to adjust the comments to state this occurs when the pattern doesn't match any files. You can keep the reference saying it may also occur on other servers normally but I just want to be clear when it normally happens in the code.

henrik-nil-acc:
Better now?
Thanks for the reviews, I have a few more similar fixes, will open PRs in a slow pace :)

jborean93 (approved):

jborean93:
Yea looks great.

> Thanks for the reviews, I have a few more similar fixes, will open PRs in a slow pace :)

Thank you very much for testing these things out and sending the PRs. I greatly appreciate them being in small chunks as it helps to review them when I find some spare time. Let me know when you think you are nearing the end and I'll prep a new release.

PR #359, opened 2026-06-23
Clear Open client state on close-response error

henrik-nil-acc:
When receive() raised anything other than FileClosed, the
old close path propagated the exception with _connected still
True and the file_id still in session.open_table. A
subsequent close() reissued SMB CLOSE on a dead file_id, and
session.disconnect(close=True) walked the stale entry.

### DIFF
--- a/src/smbprotocol/open.py
+++ b/src/smbprotocol/open.py
@@ -1103,6 +1103,10 @@ class SMB2SetInfoResponse(Structure):
 
 
 class Open:
+    # MS-SMB2 3.2.4.1.4 ChainedFileId: 0xFF * 16 signals "use the Open
+    # from the most recent CREATE in this related-compound chain".
+    _RELATED_COMPOUND_FILE_ID = b"\xff" * 16
+
     def __init__(self, tree, name):
         """
         [MS-SMB2] v53.0 2017-09-15
@@ -1126,10 +1130,7 @@ class Open:
         self.file_attributes = None
 
         # properties used privately
-        # set to { 0xFFFFFFFFFFFFFFFF, 0xFFFFFFFFFFFFFFFF } to allow message
-        # compounding with the open as the first message, once opened this
-        # will be overwritten by the open response
-        self.file_id = b"\xff" * 16
+        self.file_id = self._RELATED_COMPOUND_FILE_ID
         self.tree_connect = tree
         self.connection = tree.session.connection
         self.oplock_level = None
--- a/src/smbprotocol/open.py
+++ b/src/smbprotocol/open.py
@@ -1597,28 +1597,29 @@ class Open:
         try:
             response = self.connection.receive(request)
         except FileClosed:
+            # already closed server-side
+            return
+        else:
+            c_resp = SMB2CloseResponse()
+            c_resp.unpack(response["data"].get_value())
+            log.debug(c_resp)
+
+            # update the attributes if requested
+            close_request = SMB2CloseRequest()
+            close_request.unpack(request.message["data"].get_value())
+            if close_request["flags"].has_flag(CloseFlags.SMB2_CLOSE_FLAG_POSTQUERY_ATTRIB):
+                self.creation_time = c_resp["creation_time"].get_value()
+                self.last_access_time = c_resp["last_access_time"].get_value()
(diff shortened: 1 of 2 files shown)


src/smbprotocol/open.py
```diff
@@ -1596,28 +1597,31 @@ def _close_response(self, request):
...
+                self.allocation_size = c_resp["allocation_size"].get_value()
+                self.end_of_file = c_resp["end_of_file"].get_value()
+                self.file_attributes = c_resp["file_attributes"].get_value()
+            return c_resp
+        finally:
+            # Entry may be absent: ChainedFileId sentinel was never
+            # registered, or a prior close already popped it.
+            self._connected = False
+            if self.file_id != self._RELATED_COMPOUND_FILE_ID:
+                self.tree_connect.session.open_table.pop(self.file_id, None)
```
jborean93: Why do we only pop it if it's not the compound file id? Wouldn't the compound id be present in the `open_table` causing it to be explicitly closed if the session is closed?

jborean93 (commented):
Is there a real life scenario we can test with rather than doing it with the mocks? I definitely prefer proper integration tests over unit tests that are just testing the shape of the code.

src/smbprotocol/open.py
henrik-nil-acc: I tweaked the change a bit. Better?

henrik-nil-acc:
> Is there a real life scenario we can test with rather than doing it with the mocks? I definitely prefer proper integration tests over unit tests that are just testing the shape of the code.

Not very easy since it's a race condition.

This works against samba:
Point `SMB_SERVER`/`SMB_SHARE`/`SMB_USER`/`SMB_PASSWORD` at a writable share:

```python
import gc, multiprocessing as mp, os, warnings

SERVER = os.environ["SMB_SERVER"]
SHARE = os.environ.get("SMB_SHARE", "share")
PORT = int(os.environ.get("SMB_PORT", "445"))
USER = os.environ.get("SMB_USER")
PASSWORD = os.environ.get("SMB_PASSWORD")
ROOT = rf"\\{SERVER}\{SHARE}\repro-359"
DIRS, FILES_PER_DIR, PROCESSES = 20, 50, 12


def _populate():
    from smbclient import makedirs, open_file, register_session
    from smbclient.shutil import rmtree
    register_session(SERVER, username=USER, password=PASSWORD, port=PORT)
    rmtree(ROOT, ignore_errors=True, port=PORT)
    makedirs(ROOT, exist_ok=True, port=PORT)
    for d in range(DIRS):
        sub = rf"{ROOT}\dir-{d:04d}"
        makedirs(sub, exist_ok=True, port=PORT)
        for f in range(FILES_PER_DIR):
            with open_file(rf"{sub}\file-{f:04d}.txt", mode="w", port=PORT) as fd:
                fd.write("x")


def _worker(queue):
    gc.set_threshold(1, 1, 1)
    count = 0

    def _show(message, category, *args, **kwargs):
        nonlocal count
        if category is ResourceWarning and "unclosed SMB handle" in str(message):
            count += 1

    warnings.simplefilter("always")
    warnings.showwarning = _show
    from smbclient import register_session
    from smbclient.shutil import rmtree
    register_session(SERVER, username=USER, password=PASSWORD, port=PORT)
    rmtree(ROOT, ignore_errors=True, port=PORT)
    gc.collect()
    queue.put(count)


def main():
    _populate()
    ctx = mp.get_context("spawn")
    queue = ctx.Queue()
    procs = [ctx.Process(target=_worker, args=(queue,)) for _ in range(PROCESSES)]
    for p in procs:
        p.start()
    total = sum(queue.get() for _ in procs)
    for p in procs:
        p.join()
    print(f"unclosed SMB handle warnings: {total}")


if __name__ == "__main__":
    main()
```

```
$ PYTHONWARNINGS=always python repro_359.py   # master
unclosed SMB handle warnings: 171
$ PYTHONWARNINGS=always python repro_359.py   # this PR
unclosed SMB handle warnings: 0
```

jborean93 (approved):

jborean93:
> Not very easy since it's a race condition.

That's fair enough thanks for clarifying.

PR #361, opened 2026-06-25
Surface SMB tree setup errors as OSError

henrik-nil-acc:
A raw SMBResponseException from SMBRawIO setup did not flow through rmtree's ignore_errors and onerror hooks. The same gap affected remove, rmdir, and scandir for callers using smbclient as a drop-in for os and os.path.

The transaction layer already translated these errors for operations wrapped in SMBFileTransaction.


jborean93 (approved):
I'll have to remember to explicit mark this in the changelog as it might catch people who are already ignoring the old exception type but this is still better to align the behaviours between how the transaction APIs work vs the standalone opens. Plus it is also nice that the `smbclient` stops re-raising `smbprotocol` defined errors directly.

PR #362, opened 2026-06-25
Move dev dependencies into pyproject.toml

henrik-nil-acc:
The requirements file was the only dependency list outside the project metadata. cryptography and pyspnego only duplicated runtime dependencies, and build is a release tool CI installs separately.


jborean93 (approved):
I'm probably move this to `uv` at some point but this is good enough.

PR #363, opened 2026-06-25
Return context-manager iterator from scandir()

henrik-nil-acc:
Abandoning scandir() iteration before exhaustion left the SMB directory handle open until GC finalized the generator, and callers had no idiomatic way to release it sooner. Stdlib os.scandir() solves this by returning a context-manager iterator.


jborean93 (approved):
Nice, I realised I made this mistake after I did the initial release but didn't think I could actually solve it without making a breaking change. Glad to see I was wrong.

PR #364, opened 2026-06-26
Switch from black and isort to ruff

henrik-nil-acc:
black and isort are two tools, two configs, and two pinned
versions to keep in step for what ruff does in one.


henrik-nil-acc:
Maybe make a release before this is merged?

jborean93:
I've pushed out v1.17.0 just now, thank you very much for all your changes and bugfixes so far. I appreciate that this consolidates everything into 1 tool but for now I prefer to keep things as they are. Most of my other projects use `black` and `isort` and I've had enough time to get a handle on how I can use `ruff` instead and what rules should or should not be enabled.

Maybe I'll revisit this in the future but for now wish to keep this as it is today.

Issue #326, opened 2025-10-02
skip_dfs=True doesn't skip DFS


p1-dta:
I have an issue with DFS with a specific server.

The error: 
> smbprotocol.exceptions.SMBOSError: [Error 0] [NtStatus 0xc000026d] Unknown NtStatus error returned 'STATUS_DFS_UNAVAILABLE': <path>

My code:

```python
import smbclient

smbclient.ClientConfig(username=smb_username, password=smb_pwd, skip_dfs=True)
smbclient.listdir("<path>")
```
The code I showed works for other servers, with and without skip_dfs.

Despite `skip_dfs=True`, it seems it still use it. I am clueless.

PR #368, opened 2026-09-16
Make skip_dfs also disable DFS on the tree connect

tiptonsc:
- `ClientConfig(skip_dfs=True)` is documented as ["skip using any DFS referral checks and treat any path as a normal path"](https://github.com/jborean93/smbprotocol/blob/55d8294a3b2a863e66c4cc207547802f4314dec7/src/smbclient/_pool.py#L76-L78), but it currently has no effect on whether a share is treated as a DFS share. Its only read is inside the `domain_controller` setter, which `__init__` bypasses by assigning `self._domain_controller` directly.
    - This fixes situations where a server advertises `SMB2_SHARE_CAP_DFS` but returns `STATUS_DFS_UNAVAILABLE` for a referral beneath it, causing `smbclient.rename` to fail.
    - Possibly the same underlying cause as issue #326

- `TreeConnect.connect` adds `use_dfs=True`, so `is_dfs_share` is `use_dfs and <capability>` rather than being based on only the capability.
- `get_smb_tree` passes `use_dfs=not client_config.skip_dfs`, and short-circuits the `BadNetworkName` → root-referral fallback when `skip_dfs` is set.

Defaults are unchanged, so existing behavior is identical.


jborean93 (approved):

jborean93:
Thanks for implementing this and adding some tests.

